PR #11, opened 2025-01-17
Implement dirac metadata via hints for CWL

mexanick:
- Add DIRAC metadata schema to the project
- Add query_params and description hints to the job model
- Provide a dummy CWL example which allows to test metadata parameters setting, updating and propagation

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,6 +33,7 @@ jobs:
       - name: Set up environment
         run: |
           pip install .
+          pip install types-jsonschema types-PyYAML  # Install type stubs
       - name: Run mypy
         run: |
           mypy src
--- a/src/dirac_cwl_proto/job/__init__.py
+++ b/src/dirac_cwl_proto/job/__init__.py
@@ -46,10 +46,6 @@ console = Console()
 def submit_job_client(
     task_path: str = typer.Argument(..., help="Path to the CWL file"),
     parameter_path: Optional[List[str]] = typer.Option(None, help="Path to the files containing the metadata"),
-    metadata_path: Optional[str] = typer.Option(None, help="Path to metadata file used to generate the input query"),
-    platform: Optional[str] = typer.Option(None, help="The platform required to run the job"),
-    priority: Optional[int] = typer.Option(10, help="The priority of the job"),
-    sites: Optional[List[str]] = typer.Option(None, help="The site to run the job"),
     # Specific parameter for the purpose of the prototype
     local: Optional[bool] = typer.Option(True, help="Run the job locally instead of submitting it to the router"),
 ):
@@ -76,11 +72,7 @@ def submit_job_client(
     console.print(f"\t[green]:heavy_check_mark:[/green] Task {task_path}")
 
     job_metadata = JobMetadataModel()
-    job_description = JobDescriptionModel(
-        platform=platform,
-        priority=priority,
-        sites=sites,
-    )
+    job_description = JobDescriptionModel()
     if task.hints:
         for hint in task.hints:
             hint_class = hint["class"]
--- a/src/dirac_cwl_proto/job/__init__.py
+++ b/src/dirac_cwl_proto/job/__init__.py
@@ -1,6 +1,7 @@
 """
 CLI interface to run a workflow as a job.
 """
+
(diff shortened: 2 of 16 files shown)


test/workflows/test_meta/test_meta.cwl
```diff
@@ -0,0 +1,23 @@
...
+
+# The inputs for this process: none.
+inputs: []
+# The outputs for this process: none.
+outputs: []
+
+baseCommand: ["echo", "Hello World"]
+
+$namespaces:
+  dirac: "./schemas/dirac-metadata.yaml#" #we place a copy locally, but it should be published
```
mr-c, line 13: Typically `$namespaces` should be used to expand a prefix like `direct` into a public URI

You can use `$schemas` if you want to link to the exact schema (in RDF-XML format) that was used. 

https://www.commonwl.org/v1.2/SchemaSalad.html#Explicit_context

aldbr: Thanks a lot for the feedback!

mexanick: @mr-c thank you for the suggestion! I want to shadow our DIRAC-specific hints, and I would like to have native schema validation to check whether hint is having a correct form. Is the use of RDF schema the only way to accomplish this?

.github/workflows/main.yml
```diff
@@ -33,6 +33,7 @@ jobs:
       - name: Set up environment
         run: |
           pip install .
+          pip install types-jsonschema types-PyYAML  # Install type stubs
```
aldbr, line 36: These would also need to be added to the `.precommit.config.yaml` : https://github.com/aldbr/dirac-cwl-proto/blob/d513b1d204fab368167deddd01cbb5866d400200/.pre-commit-config.yaml#L30

src/dirac_cwl_proto/job/__init__.py
```diff
@@ -67,39 +60,54 @@ def submit_job_client(
...
     )
+    for hint in task.hints:
+        hint_class = hint["class"]
+        hint_stripped = {k: v for k, v in hint.items() if k != "class"}
+        if hint_class == "dirac:metadata":
+            console.print(f"Update metadata with:\n{hint_stripped}")
+            job_metadata = job_metadata.copy(update=hint_stripped)
+            continue
+        if hint_class == "dirac:description":
+            job_description = job_description.copy(update=hint_stripped)
```
aldbr: Correct me if I am wrong: 
- you kept the `job_metadata` and `job_description` input parameters of the CLI just to avoid breaking the current workflows and tests, right?
- but we actually don't need them anymore since this is now directly supported within CWL

src/dirac_cwl_proto/job/__init__.py
```diff
@@ -67,39 +60,54 @@ def submit_job_client(
...
 
     parameters = []
     if parameter_path:
         for parameter_p in parameter_path:
             parameter = load_inputfile(parameter_p)
 
+            overrides = parameter.pop("cwltool:overrides", {})
+            if overrides:
+                if len(overrides) > 1:
+                    # QUESTION: What's the best way to handle a ValueError? Raising it or logging and exiting wiht 1:
```
aldbr, line 100: As we are supposed to be in the CLI used by the end user here, I would actually return a "friendly" output to him/her, and so I would log and exit with a status code != 0

aldbr (changes requested):

src/dirac_cwl_proto/job/__init__.py
mexanick: I believe we still need them, as the information in hints is our custom extension. CWL reference runner will just log a warning that it encountered unknown hints. An alternative would be to implement custom hints.

aldbr (commented):
I see that you have not applied your logic to the production endpoint in the PR. Do you want to do it in a separate PR later?

test/workflows/test_meta/test_meta.cwl
mr-c: > Is the use of RDF schema the only way to accomplish this?

https://www.commonwl.org/v1.2/SchemaSalad.html#Explicit_context

RDF-XML, yes

PR #12, opened 2025-01-20
feat(workflows): reference subworkflows in main workflows

aldbr:
- [x] tests should pass
- [x] add a test where the subworkflow does not exist and make sure it is correctly handled
- [x] add a test where the subworkflow references the parent workflow (circular dependency)
- [x] add a test where the subworkflow references itself


arrabito:
Thank you @aldbr for this MR. 
All seems good to me.

arrabito (approved):

PR #13, opened 2025-02-27
example of 2 dependent workflows each composed of 2 independent cmdlinetool

natthan-pigoux:
Example which should cover the "1 Workflow with 2 dependent Workflow, where each of them is composed of 2 independent CommandLineTools" case:
(o  o)
  |
(o  o)

There is two scripts:
- `random-data-gen` : generate random gaussian data
- `gaussian-fit`: fit the data with gaussian param (mean and std dev)

There is two CommandLineTools, one running the data generation and the other the gaussian fit scripts. 
They are then used in the "transformation" workflow, to generate two different data files and to fit each data files seperatly.
Finally, there is a `main-workflow.cwl` which run the two sub-workflows composed of the two CLT.

The use case is a bit hacked, but the principle is here. I can make a more realistic example if needed.


test/test_workflows.py
```diff
@@ -131,6 +131,28 @@ def _cleanup():
...
+                "test/workflows/gaussian_fit/type_dependencies/transformation/inputs-data-generation.yaml"
+            ],
+        ),
+        # Gaussian fit workflow
+        (
+            "test/workflows/gaussian_fit/gaussian_fit/gaussian-fit-workflow.cwl",
+            [
+                "test/workflows/gaussian_fit/type_dependencies/transformation/inputs-gaussian-fit.yaml"
+            ],
+        ),
```
aldbr, line 155: I see that you are only testing the workflow as a job, is that expected?

aldbr (commented):
The overall structure looks good to me.
I have a question though: I see you are passing some input data to the main workflow. While it would work fine as a job, and as a transformation, it would probably don't work as a production (because here we have no way of passing input data). Is that expected?

I guess this is the reason why you are only testing the workflow as a job, isn't it?

test/test_workflows.py
natthan-pigoux: No... the tests for the workflows are in the right place. I correct that right away. Thanks for noticing.

natthan-pigoux: I finally made a lot of changes, and was able to make job, transformation and production examples working.

natthan-pigoux: I'm passing input data only for the transformation. So it should work with the production.

natthan-pigoux:
@aldbr , I added job, transformation and production examples, which should work now. Can you check I did not cheat somewhere? :)

natthan-pigoux:
The CWL Preview for the production gives:

Screenshot from 2025-03-19 12-09-00

natthan-pigoux:
@aldbr , do I merge this?

aldbr (approved):

PR #15, opened 2025-03-18
test: allow multiple destinations for blocking transformation test

natthan-pigoux:
To run blocking transformations, instead of specifying a source inputs array and a single destination. We specify now a "dest_source" dict with `{destination: [inputs, ..]}`


aldbr:
Ah now I understand what you meant! Perfect, thanks :slightly_smiling_face:

aldbr (approved):

PR #34, opened 2025-09-25
Check post-process returned value for job success

### LINKED ISSUE #33: Post-process return value is ignored when running a job
In the run_job function, the value returned by the post process of the job wrapper is ignored. This means the job can still report a success even though the post-process failed.

https://github.com/aldbr/dirac-cwl-proto/blob/0a7ab0154feadd641e17fdcabf3f0b83105fd4df/src/dirac_cwl_proto/job/__init__.py#L396-L408

Checking for the post-process success makes 4 tests fail
```
=========================== short test summary info ============================
FAILED test/test_workflows.py::test_run_nonblocking_transformation_success[test/workflows/lhcb/lhcbsimulate.cwl-test/workflows/lhcb/type_dependencies/transformation/metadata-lhcb_simulate.yaml]
FAILED test/test_workflows.py::test_run_nonblocking_transformation_success[test/workflows/gaussian_fit/data_generation/data-generation.cwl-test/workflows/gaussian_fit/type_dependencies/transformation/inputs-data-generation.yaml]
FAILED test/test_workflows.py::test_run_blocking_transformation_success[test/workflows/lhcb/lhcbreconstruct.cwl-test/workflows/lhcb/type_dependencies/transformation/metadata-lhcb_reconstruct.yaml-destination_source_input_data1]
FAILED test/test_workflows.py::test_run_simple_production_success[test/workflows/lhcb/description.cwl-test/workflows/lhcb/type_dependencies/production/metadata-lhcb_complete.yaml]
============= 4 failed, 165 passed, 1 warning in 244.34s (0:04:04) =============
```

Loxeris:
Aims to fix #33 and the failed tests.

The lhcb plugin had some incorrect function calls, fixing it only leaves the gaussian-fit test failing.
It seems the test for the gaussian-fit data generation was using the wrong cwl task but there was still an error after changing that.
The job's outputs were using default filenames instead of the ones defined in the configuration. I tried to fix that but I'm not 100% sure it's working as intended in all scenarios, even though the tests are now all passing.

### DIFF
--- a/src/dirac_cwl_proto/transformation/__init__.py
+++ b/src/dirac_cwl_proto/transformation/__init__.py
@@ -189,13 +189,6 @@ def submit_transformation_router(transformation: TransformationSubmissionModel) 
         job_model_params = _generate_job_model_parameter(input_data_dict)
         logger.info("Input data for the transformation retrieved!")
 
-    elif transformation.execution_hooks.configuration:
-        job_model_params.append(
-            JobInputModel(
-                sandbox=None, cwl=transformation.execution_hooks.configuration
-            )
-        )
-
     logger.info("Building the jobs...")
     jobs = JobSubmissionModel(
         task=transformation.task,
--- a/test/workflows/gaussian_fit/data_generation/data-generation-workflow.cwl
+++ b/test/workflows/gaussian_fit/data_generation/data-generation-workflow.cwl
@@ -7,10 +7,10 @@ doc: >
 inputs:
   output_file_name_1:
     type: string
-    default: "data1.txt"
+    default: "data-gen1.txt"
   output_file_name_2:
     type: string
-    default: "data2.txt"
+    default: "data-gen2.txt"
 
 outputs:
   data1:


src/dirac_cwl_proto/transformation/__init__.py
```diff
@@ -186,6 +186,13 @@ def submit_transformation_router(transformation: TransformationSubmissionModel)
         job_model_params = _generate_job_model_parameter(input_data_dict)
         logger.info("Input data for the transformation retrieved!")
 
+    elif transformation.execution_hooks.configuration:
+        job_model_params.append(
+            JobInputModel(
+                sandbox=None, cwl=transformation.execution_hooks.configuration
+            )
+        )
+
```
aldbr, line 195: Does these lines work?
I might be wrong but I think `JobInputModel.cwl` is expected to be a CWL input, whereas the `transformation.execution_hooks.configuration` is expected to be a dict mapping an input key to a given value.
They look very similar but they are not exactly the same, and I agree this is not clear within the code 😅 

I am not sure to understand why you need them? Can you explain the issue you got leading you to this block of code?

src/dirac_cwl_proto/job/__init__.py
```diff
@@ -396,15 +396,16 @@ def run_job(job: JobSubmissionModel) -> bool:
...
             result.stderr,
             job_path,
             runtime_metadata,
-        )
-        logger.info("Task post-processed successfully!")
-        return True
+        ):
+            logger.info("Task post-processed successfully!")
+            return True
+        return False
```
aldbr, line 408: Could you add a `logger.error("Failed to post-process Task")` here?

aldbr (commented):
Well spotted, thanks @Loxeris 🙂

src/dirac_cwl_proto/transformation/__init__.py
aldbr: Note: we could potentially add a `TypedDict` for the CWL input to distinct them.

PR #35, opened 2025-09-25
Fix semantics

### LINKED ISSUE #20: About using the right semantics
In the current prototype, we have multiple overlapping concepts around metadata models, hints, and submission models.
Some terms (like `MetadataModel`) may be misleading or overloaded (e.g. `metadata` already has a specific meaning in DiracX for scheduling and monitoring).

This issue collects the main pain points and proposals (sorry it's a bit messy), with the goal of:
- Clarifying naming conventions and responsibilities of each class/interface.
- Aligning hint semantics with what’s already common in other workflow management systems.
- Deciding whether to merge or separate hints for data staging and scheduling.
- Cleaning up submission models to remove redundancy.

### Metadata models

Will be used by the `JobWrapper` to "inject" specific actions depending on the VO and the "job type":
- hooks to pre-process and post-process the CWL
- data catalog plugin to get specific inputs and store the outputs

**Potential problem with the current situation with the `MetadataModel`**

`Metadata` in the context of `diracx` already refers to data used to schedule the jobs (`sites`, `priority`), as well as data obtained from the execution of the job (`wallclock time`, `worker node`). 

```mermaid
classDiagram
  direction LR

  class MetadataProcessor {
    <<abstract>>
    +pre_process()
    +post_process()
  }

  class DataCatalogInterface {
    <<abstract>>
    +get_input_query() 
    +get_output_query()
    +store_output()
  }

  class BaseMetadataModel {
    <<pydantic>>
    +vo
    +version
    +description
    +pre_process()
    +post_process()
    +get_input_query() 
    +get_output_query()
    +store_output()
  }

  BaseMetadataModel <|-- MetadataProcessor
  BaseMetadataModel <|-- DataCatalogInterface
```

**Proposed Solution**
- `ExecutionHooks`: often refers to custom code snippets that run before or after a specific action occurs.
- `DataCatalogBasePlugin`: still not entirely sure about it. But it indicates you can adapt the data source/target I think.
- `TaskRuntimeBasePlugin`: still not entirely convinced. But it also indicates that one can adapt the task runtime.

Note: we switch from inheritance to composition here.
Note: I used `Plugin` instead of `Adapter` because `Adapter` generally refers to an oriented object design pattern, and we are not implementing it here (http://www.goprod.bouhours.net/?page=pattern&pat_id=1)
Note: input and output sandbox will be managed by the `JobWrapper` because it's generic. I am not sure about the LFNs.

```mermaid
classDiagram
  direction LR

  class ExecutionHooks {
    <<abstract>>
    +pre_process()
    +post_process()
  }

  class DataCatalogBasePlugin {
    <<abstract>>
    +get_input_query() 
    +get_output_query()
    +store_output()
  }

  class TaskRuntimeBasePlugin {
    <<pydantic>>
    +vo
    +version
    +description
    +pre_process()
    +post_process()
    +get_input_query() 
    +get_output_query()
    +store_output()
  }

  TaskRuntimeBasePlugin o-- ExecutionHooks
  TaskRuntimeBasePlugin o-- DataCatalogBasePlugin
```

---

### Dirac-specific hints

**Potential problem with the current hints**

We have 2 different hints:
- `dirac:data-management`:

  ```yml
  dirac:data-management:
    metadata_class: "TaskWithMetadataQueryPlugin"
    query_params:
    site: "Paranal"
  ```

- `dirac:job-execution`:

  ```yml
  dirac:job-execution:
    platform: x86_64
    priority: 5
    sites: ["CSCS", "PIC"] 
  ```

First of all, should we merge them together?
- On the one hand, it would be easier to manage, one would just have to add 1 section to submit a job to Dirac.
- On the other hand, I think the `dirac:data-management` hint will only be used by admins submitting production jobs or powerful users, so most of the user will not have to care about it (unless we add everything related to data management inside).

Is the semantics correct here?
- `dirac:data-management`: does not integrate input/output sandboxes and lfns currently. They are just part of the `JobParameters` and not defined in CWL at all. 
  One thing to take into account though, we want to submit multiple jobs in 1 request, by adding 1 workflow + different CWL parameters (e.g. submitting 1 workflow with 5 cwl parameters will generate 5 jobs). But I think the query parameters will be the same for all the jobs, so putting them together will create duplication of query parameters.
If we do add everything related to input/output data together, we could also potentially name it `dirac:data-staging`.
- `dirac:job-execution`: I would probably call it `dirac:scheduling` because it will contain dirac-specific attributes to schedule the job on a suitable worker node I think (by the way we should check whether a parameter like `platform` is not already included in CWL).
  Also I have a question: in a production, we have different steps, and like the query parameters, I guess we should also have one `dirac:job-execution` hint per step, am I wrong?

  Note: input sandbox are not defined in the hints currently, I am not 100% we want to include them though. There are 2 different phases: i) the user provides a CWL and some parameter and hints to the `diracx` CLI; ii) the CLI creates the input sandboxes and links them to the json request sent to the server. Here we just define the hints needed to interact with the CLI, but we could include hints to interact with the `diracx` services.

How to process the `metadata_class`?
-  It was previously defined as `job_type`, and still is in `diracx`. It is used to instantiate the right `MetadataModel` (or `TaskRuntimePlugin`), but it is also used as a scheduling attribute (e.g. we limit the execution of n jobs of a given type on site A). 

Now about the classes behind the hints:
- `DataManager`  is behind `dirac:data-management`.
  I am not entirely sure to understand the purpose of `vo` here. Is that supposed to be used to extend it for vo-specific fields?
  I wonder whether it makes sense for `DataManager` to return a `BaseMetadataModel`: I would say yes in the current context, because input/output sandboxes and lfns are not treated here, and `metadata_class` (job type) is not part of the `dirac:job-execution` hint.
  There is no `ProductionDataManager`, is that expected? Finding a hint that would work for both a production and transformation/job is probably not easy.
- `JobExecutor` is behind `dirac:job-execution`.
  Both `DataManager` and `JobExecutor` could implement a `CWLHint` interface actually, because they both have `from_cwl_hints()` (unless we merge them).
  I am not sure to understand the purpose of `TaskDescriptor` as it is composed of a single method `from_hints()`, which just calls `JobExecutor.from_cwl_hints()`. Unless I am missing something, I would just remove it.

```mermaid
classDiagram
  direction LR

 class DataManager {
    <<pydantic>>
    +vo
    +version
    +metadata_class
    +query_params
    +model_copy()
    +to_runtime()
    +from_cwl_hints()
  }
  class TransformationDataManager {
    <<pydantic>>
    +group_size
  }
  DataManager <|-- TransformationDataManager

  class JobExecutor {
    <<pydantic>>
    +platform
    +priority
    +sites
    +from_cwl_hints()
  }
  class TaskDescriptionModel {
    <<pydantic>>
    +from_hints()
  }
  JobExecutor <|-- TaskDescriptionModel
```

**Proposed Solution**

We would likely need to have different hints for the productions actually, as they need to support of list of `dirac:job-execution` and `dirac:data-management` fields (1 per transformation).

```mermaid
classDiagram
  direction TB

 class CWLHint {
   <<abstract>>
    +version
    +from_cwl_hints()
 }

 class DataStagingCWLHint {
    <<pydantic>>
    +metadata_class
    +query_params
    +model_copy()
  }
  class StepDataStagingCWLHint {
    <<pydantic>>
    +group_size 
  }
class WorkflowDataStagingCWLHint {
    <<pydantic>>
  }
  
  class SchedulingCWLHint {
    <<pydantic>>
    +platform
    +priority
    +sites
  }
  class WorkflowShedulingCWLHint {
    <<pydantic>>
  }

  CWLHint <|-- DataStagingCWLHint
  CWLHint <|-- SchedulingCWLHint
  
  DataStagingCWLHint <|-- StepDataStagingCWLHint

  WorkflowShedulingCWLHint o-- "*" SchedulingCWLHint
  WorkflowDataStagingCWLHint o-- "*" StepDataStagingCWLHint

```

---

### Submission models

**Potential problem with the current models**

- As said before, it looks like `TaskDescriptor` could disappear.
- `TransformationMetadataModel` could potentially also be replaced with `TransformationDataManager`?
- `ProductionStepMetadataModel` could also disappear.

```mermaid
classDiagram
  direction TB

  class JobParameterModel {
    <<pydantic>>
  }

  class JobSubmissionModel {
    <<pydantic>>
  }
  JobSubmissionModel o-- "1" TaskDescriptionModel : description
  JobSubmissionModel o-- "1" DataManager : metadata
  JobSubmissionModel o-- "*" JobParameterModel : parameters

  class TransformationMetadataModel {
    <<pydantic>>
    +group_size : Dict[str,int]|None
  }
  DataManager <|-- TransformationMetadataModel

  class TransformationSubmissionModel {
    <<pydantic>>
  }
  TransformationSubmissionModel o-- "1" TransformationMetadataModel : metadata
  TransformationSubmissionModel o-- "1" TaskDescriptionModel : description

  class ProductionStepMetadataModel {
    <<pydantic>>
  }
  ProductionStepMetadataModel o-- "1" TaskDescriptionModel : description
  ProductionStepMetadataModel o-- "1" TransformationMetadataModel : metadata

  class ProductionSubmissionModel {
    <<pydantic>>
  }
  ProductionSubmissionModel o-- ProductionStepMetadataModel : steps_metadata
```

**Proposed Solution**

```mermaid
classDiagram
  direction LR

  class JobParameterModel {
    <<pydantic>>
  }

  class JobSubmissionModel {
    <<pydantic>>
  }
  JobSubmissionModel o-- "1" SchedulingCWLHint 
  JobSubmissionModel o-- "1" DataStagingCWLHint
  JobSubmissionModel o-- "*" JobParameterModel

  class TransformationSubmissionModel {
    <<pydantic>>
  }
  TransformationSubmissionModel o-- "1" StepDataStagingCWLHint
  TransformationSubmissionModel o-- "1" SchedulingCWLHint

  class ProductionSubmissionModel {
    <<pydantic>>
  }
  ProductionSubmissionModel o-- "*" WorkflowDataStagingCWLHint
  ProductionSubmissionModel o-- "*" WorkflowSchedulingCWLHint
```

mexanick:
Renaming "metadata" module and MetadataPluginRegistry to "execution_hooks" and "ExecutionHooksRegistry".

 The core classes are now looking like this (zoomable!):

image

Closes #20


test/test_metadata_plugins_core.py
aldbr, line 1: Shouldn't we delete this file?
IIUC, it has been replaced with `test_plugins_lhcb.py` right?
Same for `test_plugins_core.py` (`test_metadata_core.py`) and `test_execution_hooks_registry.py` (`test_metadata_registry.py`)

test/test_plugins_lhcb.py
```diff
@@ -0,0 +1,380 @@
+"""
+Tests for the LHCb experiment metadata plugins.
```
aldbr, line 2: Minor detail: Here we could also rename `metadata`. Same for the other test files.

aldbr (commented):
A few minor comments, other than that it looks good to me.

test/test_plugins_lhcb.py
mexanick: Good catch! Done.

arrabito:
@mexanick yes I agree, it's better to treat this aspect and the general design of the DataCatalogInterface in a separate issue.

PR #36, opened 2025-09-26
Temporary disable non-core plugins from updates and testing

mexanick:
Removing the non-core plugins non-necessary core plugins and corresponding tests to reduce the codebase.


test/test_workflows.py
```diff
@@ -624,18 +457,6 @@ def test_run_simple_production_success(cli_runner, cleanup, cwl_file, metadata):
...
-            "test/workflows/mandelbrot/description.cwl",
-            "test/workflows/mandelbrot/type_dependencies/production/malformed-wrong-stepname_metadata-mandelbrot_complete.yaml",
-            "Thefollowingstepsaremissingfromthetaskworkflow:{'this-step-doesnot-exist'}",
-        ),
-        # The metadata has an unexistent type
-        (
-            "test/workflows/mandelbrot/description.cwl",
-            "test/workflows/mandelbrot/type_dependencies/production/malformed-nonexisting-type_metadata-mandelbrot_complete.yaml",
-            "Unknownexecutionhooksplugin:'MandelBrotDoesNotExist'",
-        ),
```
aldbr: Is this tested somewhere else? Don't we want to keep such tests?

test/test_workflows.py
```diff
@@ -540,35 +402,6 @@ def test_run_transformation_validation_failure(
         # --- Crypto example ---
         # Complete
         ("test/workflows/crypto/description.cwl", None),
-        # --- Pi example ---
-        # There is no input expected
-        (
-            "test/workflows/pi/description.cwl",
-            "test/workflows/pi/type_dependencies/production/metadata-pi_complete.yaml",
-        ),
```
aldbr: Don't we want to keep at least `Pi`? Because from what I see you didn't keep any `Workflow`, is that expected?

aldbr: Keeping a small number of workflows can help us seeing concretely what we are doing.
Unless I miss something?

test/test_workflows.py
mexanick: The plugin system is tested in the unit tests, but end-to-end is indeed removed. I would add it back, but using simpler workflows. Do you want me to add it in the PR, or is it better to do in the separate?

test/test_workflows.py
mexanick: we do have crypto workflows, so some small tests are still running.

mexanick:
@aldbr refactoring done, answers provided inline, let me know :)

pyproject.toml
```diff
@@ -42,13 +42,8 @@ testing = ["pytest>=6", "pytest-mock"]
 
 [project.scripts]
 dirac-cwl = "dirac_cwl_proto:app"
-pi-simulate = "dirac_cwl_proto.modules.pi_simulate:app"
-pi-simulate-v2 = "dirac_cwl_proto.modules.pi_simulate_v2:app"
-pi-gather = "dirac_cwl_proto.modules.pi_gather:app"
 crypto = "dirac_cwl_proto.modules.crypto:app"
-lhcb-app = "dirac_cwl_proto.modules.lhcb_app:app"
 random-data-gen = "dirac_cwl_proto.modules.random_data_gen:app"
```
aldbr: From what I can see you deleted this module
```suggestion
```

aldbr (approved):

Issue #31, opened 2025-09-22
Interpret the hints from the server only?


aldbr:
The client can just add them to the CWL if they are separated (`pack`)

---

- Also, in the `SubmissionModels`, I have the feeling it would be cleaner to read the hints just once, in the service. They would come either integrated in the workflow or as a separate file that would be merged within the workflow in the client code (`pack`). 
- So the `SubmissionModels` would not carry the hints: they would be in the CWL. The advantage is that a given user can provide a CWL through the REST interface in a more "standard" way, not having to specify the hints separately in the JSON blob.
- And then they would be read by the services (as the clients are doing right now).

What's your opinion?

_Originally posted by @aldbr in https://github.com/aldbr/dirac-cwl-proto/issues/23#issuecomment-3311418538_

aldbr:
> [@aldbr](https://github.com/aldbr) If I understand correctly, you want to remove those from the `TransformationSubmissionModel`:
> 
> [dirac-cwl-proto/src/dirac_cwl_proto/submission_models.py](https://github.com/aldbr/dirac-cwl-proto/blob/e97b83d0d65a46042fa9b9fbb725b5bbed498665/src/dirac_cwl_proto/submission_models.py#L76-L77)
> 
> Lines 76 to 77 in [e97b83d](/aldbr/dirac-cwl-proto/commit/e97b83d0d65a46042fa9b9fbb725b5bbed498665)
> 
>  execution_hooks: TransformationExecutionHooksHint 
>  scheduling: SchedulingHint 
> And then, in the client, it would only be something like this:
> 
> ```
>  transformation = TransformationSubmissionModel(task=task)
> ```
> 
> And in the router, when asking for the hints, we'd call: `transformation_execution_hooks, transformation_scheduling = extract_dirac_hints(transformation.task)` Instead of: `transformation.execution_hooks` and `transformation.scheduling`?

Yes exactly!
For context, before recent changes the Dirac-specific hints (like `SchedulingHint` and `ExecutionHooksHint`) were expressed outside of the CWL workflow, this is why they are part of the pydantic models and separated from the `task`.
Now, they are meant to be part of the CWL, and so we don't want to let the users expressing them both within the CWL workflow and withing the JSON blob sent to the server (outside of the CWL).
We just haven't had the time to remove them from the pydantic models.

Is it clear? 😅 


> Does the `pack` part means that we combine those hints data to the ones we already have in the CWL?:

[...]

PR #38, opened 2025-10-07
Updated Transformation hints

### LINKED ISSUE #30: Streamline the Transformation CLI
- These parameters should of course be taken from the hints and not from the CLI, as we do for the jobs: https://github.com/aldbr/dirac-cwl-proto/blob/5d68b8e8c2afb425c49e7f4494131c6f3a9ae65a/src/dirac_cwl_proto/transformation/__init__.py#L44-L52

_Originally posted by @aldbr in https://github.com/aldbr/dirac-cwl-proto/issues/23#issuecomment-3311418538_

Stellatsuu:
Merge request closes #30 for now. 
Drafted because #31 might be added to this MR since it's the same subject.

### DIFF
--- a/src/dirac_cwl_proto/transformation/__init__.py
+++ b/src/dirac_cwl_proto/transformation/__init__.py
@@ -40,16 +40,6 @@ def submit_transformation_client(
     metadata_path: Optional[str] = typer.Option(
         None, help="Path to metadata file used to generate the input query"
     ),
-    # Dirac-specific parameters that are used by the jobs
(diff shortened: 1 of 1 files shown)


aldbr (approved):
Thanks a lot @Stellatsuu!
I think we can merge this now, I did not notice I wrote twice the same content in 2 different issues.
The rest of https://github.com/aldbr/dirac-cwl-proto/issues/31 is a bit different and can go in a different PR I would say (I removed the duplicated text).
May I merge this one? 🙂

Issue #25, opened 2025-09-22
Input/Output Sandbox/Data Management


aldbr:
We need to clarify and standardize how `diracx-cli`, `diracx-routers`, and `JobWrapper` handle **input sandboxes**, **input data**, **output sandboxes**, and **output data**.
This issue documents the current approach, proposes potential solutions, and highlights open questions.

---

### **1. Input Sandbox**

```mermaid
sequenceDiagram
  Actor user
  participant diracx-cli
  participant diracx-routers[/jobs]
  participant diracx-routers[/jobs/sandbox]
  participant job-wrapper

  user ->>+ diracx-cli: cwl workflow + cwl hints + cwl input files
  diracx-cli ->>+ diracx-routers[/jobs/sandbox]: create input sandbox + upload local files
  diracx-cli ->>+ diracx-cli: reference sandbox ids in the JSON Payload
  diracx-cli ->>+ diracx-routers[/jobs]: submit JSON Payload (see example below)
  diracx-routers[/jobs] ->>+ diracx-routers[/jobs]: generate 1 job per cwl input file
  diracx-routers[/jobs] ->>+ job-wrapper: (submit) execute
  job-wrapper ->>+ job-wrapper: pre-process - download input sandbox + run cwl workflow with inputs
```

**Example JSON Payload:**

```jsonc
{
  "payload": {
    // CWL workflow + hints
  },
  "inputs": [
    {
      "sandbox": "url-to-input-sandbox",
      "input": {
        // CWL input file content
      }
    }
  ]
}
```

**Open Question:**

* What would happen if the users provide n times the hints, and m cwl input files?
  I would naively say they can only give 1 time the hints, if they come with multiple, they should be rejected

---

### **2. Input Data**

```mermaid
sequenceDiagram
  Actor user
  participant diracx-cli
  participant diracx-routers[/jobs]
  participant job-wrapper

  user ->>+ diracx-cli: cwl workflow + cwl hints + cwl input files (containing lfns?)
  diracx-cli ->>+ diracx-routers[/jobs]: submit JSON Payload (see example below)
  diracx-routers[/jobs] ->>+ diracx-routers[/jobs]: generate 1 job per cwl input file
  diracx-routers[/jobs] ->>+ job-wrapper: (submit) execute
  job-wrapper ->>+ job-wrapper: pre-process - resolve input data (download or provide lfns/pfns mapping)  + run cwl workflow with inputs
```

**Example CWL Input:**

```yaml
[...]

PR #39, opened 2025-10-08
Input/Output Sandbox/Data Management

Loxeris:
- Add sandbox outputs to execution hooks, with a default implementation.
- Data catalogs are still defined in execution hooks, and implemented in plugins.
- LFNs outputs can be overridden in the hints.
- By default, we get the cwl outputs from the cwltool output.
- If the output has a valid output path in the DataCatalog interface, or if it's defined in the hints, we output it as a LFN, else in a sandbox. This can be easily redefined in a plugin.

See #25

### DIFF
--- a/src/dirac_cwl_proto/transformation/__init__.py
+++ b/src/dirac_cwl_proto/transformation/__init__.py
@@ -51,10 +51,7 @@ def submit_transformation_client(
     - Validate the workflow
     - Start the transformation
     """
-    if local:
-        os.environ["DIRAC_PROTO_LOCAL"] = "1"
-    else:
-        os.environ["DIRAC_PROTO_LOCAL"] = "0"
+    os.environ["DIRAC_PROTO_LOCAL"] = "0"
     # Validate the workflow
     console.print(
         "[blue]:information_source:[/blue] [bold]CLI:[/bold] Validating the transformation..."
--- a/src/dirac_cwl_proto/job/job_wrapper.py
+++ b/src/dirac_cwl_proto/job/job_wrapper.py
@@ -219,6 +219,7 @@ class JobWrapper:
             updates = self.__download_input_data(arguments, self.job_path)
             self.__update_inputs(arguments, updates)
 
+            logger.info("Preparing the parameters for cwltool...")
             parameter_dict = save(cast(Saveable, arguments.cwl))
             parameter_path = self.job_path / "parameter.cwl"
             with open(parameter_path, "w") as parameter_file:
@@ -229,15 +230,6 @@ class JobWrapper:
             return self.execution_hooks_plugin.pre_process(
                 executable, arguments, self.job_path, command
             )
-        else:  # done in execution hooks otherwise
-            if arguments:
-                # Prepare the parameters for cwltool
-                logger.info("Preparing the parameters for cwltool...")
-                parameter_dict = save(cast(Saveable, arguments.cwl))
-                parameter_path = self.job_path / "parameter.cwl"
-                with open(parameter_path, "w") as parameter_file:
-                    YAML().dump(parameter_dict, parameter_file)
-                command.append(str(parameter_path.name))
 
         return command
 
@@ -320,8 +312,8 @@ class JobWrapper:
             logger.error("Failed to post-process Task")
             return False
 
-        except Exception as e:
-            logger.exception(f"JobWrapper: Failed to execute workflow : {e}")
+        except Exception:
+            logger.exception("JobWrapper: Failed to execute workflow")
             return False
         finally:
             # Clean up
--- a/src/dirac_cwl_proto/production/__init__.py
+++ b/src/dirac_cwl_proto/production/__init__.py
@@ -54,10 +54,7 @@ def submit_production_client(
     - Start the production
     """
 
-    if local:
-        os.environ["DIRAC_PROTO_LOCAL"] = "1"
-    else:
-        os.environ["DIRAC_PROTO_LOCAL"] = "0"
+    os.environ["DIRAC_PROTO_LOCAL"] = "0"
 
     # Validate the workflow
     console.print(


src/dirac_cwl_proto/execution_hooks/core.py
```diff
@@ -182,9 +317,39 @@ def get_output_query(self, output_name: str, **kwargs: Any) -> Optional[Path]:
...
+    def get_output_type(
+        self, output_name: str, src_path: str | Path, **kwargs: Any
+    ) -> OutputType:
+        """Whether the output must be stored in a Sandbox or the Data Catalog."""
+        if output_name in self.lfns_output_overrides or self.get_output_query(
+            output_name, **kwargs
+        ):
+            return OutputType.Data_Catalog
+        else:
+            return OutputType.Sandbox
```
aldbr: I might not understand it correctly, but this logic seems very different from what we have in DIRAC.
In DIRAC, in the JDL, we specifically specify which outputs we expect in the Data Catalog (`OutputData` section) and in the Output Sandbox (`OutputSandbox` section). 
Any particular reason for not reproducing this, or something similar, here?

src/dirac_cwl_proto/execution_hooks/core.py
```diff
@@ -162,16 +264,49 @@ def pre_process(
...
+        outputs = json.loads(stdout)
+        for output, files in outputs.items():
+            if files:
+                if not isinstance(files, List):
+                    files = [files]
+                file_paths = []
+                for file in files:
+                    if file:
+                        file_paths.append(str(file["path"]))
+                outputted_files[output] = file_paths
```
aldbr: Very minor and a bit more pythonic: 
```suggestion
        outputted_files: dict[str, list[str]] = {}
        outputs = json.loads(stdout)
        
        for output, files in outputs.items():
            if not files:
                continue
                
            if not isinstance(files, List):
                files = [files]
            file_paths = []
            for file in files:
                file_paths.append(str(file["path"]))
            outputted_files[output] = file_paths
```

src/dirac_cwl_proto/execution_hooks/core.py
```diff
@@ -118,11 +202,19 @@ class ExecutionHooksBasePlugin(BaseModel):
...
+        description="Data Catalog output paths overrides", default={}
+    )
+
     # Private attribute for data catalog interface - not part of Pydantic model validation
     _data_catalog: DataCatalogInterface = PrivateAttr(
         default_factory=DummyDataCatalogInterface
     )
 
+    # Private attribute for sandbox interface - not part of Pydantic model validation
+    _sandbox_interface: SandboxInterface = PrivateAttr(default_factory=SandboxInterface)
```
aldbr: Same question with this private attribute.

aldbr (commented):
Thanks @Loxeris!
As you can see in my comments, I am not fully understanding the concepts.
Could you add more documentation please? (I know there is none at the moment... 😅)
And could you explain why the approach is so different from what we have in the JDL please?

src/dirac_cwl_proto/execution_hooks/DataManagement/Sandbox.py
```diff
@@ -0,0 +1,71 @@
+import logging
+import random
+import tarfile
+from pathlib import Path
+from typing import Optional, Sequence
+
+logger = logging.getLogger(__name__)
+
+
+class SandboxStoreClient:
```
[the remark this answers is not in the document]
aldbr: I'm just wondering, do you actually need to have a class here? Why not just some functions?

src/dirac_cwl_proto/execution_hooks/DataManagement/Sandbox.py
```diff
@@ -0,0 +1,71 @@
...
+    """Class used for sandbox operation"""
+
+    def uploadFilesAsSandbox(
```
aldbr: In python, methods/functions should be lower case with `_` (don't look at what we are doing in DIRAC 😅 ):

```suggestion
    def upload_files_as_sandbox(
```

src/dirac_cwl_proto/execution_hooks/DataManagement/FileCatalog.py
aldbr: Same here: `file_catalog.py`

src/dirac_cwl_proto/execution_hooks/DataManagement/DataManager.py
aldbr: And here: `data_manager.py`

src/dirac_cwl_proto/execution_hooks/DataManagement/DataManager.py
```diff
@@ -0,0 +1,87 @@
...
+        if file_catalog == "LocalFileCatalog":
+            self.file_catalog = LocalFileCatalog()
+            self.base_storage_path = "filecatalog"
+            self.storage_element = FileStorage(
+                "local", {"Path": self.base_storage_path}
+            )
+        else:
+            self.file_catalog = None
+
+    def getReplicasFromDirectory(self, directory):
```
aldbr: Same for these methods too

src/dirac_cwl_proto/execution_hooks/DataManagement/data_manager.py
```diff
@@ -0,0 +1,225 @@
+from pathlib import Path
+
+from DIRAC.Resources.Storage.FileStorage import FileStorage  # type: ignore[import-untyped]
+
+from dirac_cwl_proto.execution_hooks.DataManagement.file_catalog import LocalFileCatalog
+
+
+class DataManager:
```
mexanick: What's the reason of this redefinition instead of using the already existing `DataManager` from [`DIRAC.DataManagementSystem.Client.DataManager`](https://github.com/DIRACGrid/DIRAC/blob/732dd350391f7360fbf8d641df3591948dced329/src/DIRAC/DataManagementSystem/Client/DataManager.py#L74)?

src/dirac_cwl_proto/execution_hooks/DataManagement/file_catalog.py
```diff
@@ -0,0 +1,88 @@
+import logging
+from abc import ABC, abstractmethod
+
+logger = logging.getLogger(__name__)
+
+
+class FileCatalogInterface(ABC):
```
mexanick: Same question as above, why don't we use [existing DIRAC implementation](https://github.com/DIRACGrid/DIRAC/blob/732dd350391f7360fbf8d641df3591948dced329/src/DIRAC/Resources/Catalog/FileCatalogClient.py#L14)?

src/dirac_cwl_proto/execution_hooks/DataManagement/data_manager.py
Loxeris: I had problems when trying to use DIRAC's DataManager directly. I think some of the class dependencies requires DIRAC to be running to function properly.

src/dirac_cwl_proto/execution_hooks/DataManagement/file_catalog.py
Loxeris: Same as above, this class uses RPC, which I believe make use of DiracX/DIRAC services

src/dirac_cwl_proto/execution_hooks/data_management/data_manager.py
```diff
@@ -0,0 +1,225 @@
+from pathlib import Path
+
+from DIRAC.Resources.Storage.FileStorage import FileStorage  # type: ignore[import-untyped]
+
+from dirac_cwl_proto.execution_hooks.DataManagement.file_catalog import LocalFileCatalog
+
+
+class DataManager:
```
mexanick: now, subprocess should not break them. Could you post what issue do you have? Alternatively, "mock" in a way @aldbr proposed: define a local class, have the minimal subset of method like in DataManager, with the same names/APIs, so it can serve as a drop-in replacement.

Loxeris: I simply mock using `with patch("DIRAC.DataManagementSystem.Client.DataManager.DataManager"):`, it's present in the codes above.
Omitting the parameter `new` gives a MagicMock by default, but other values didn't change anything.

I get this error: 
```
            try:
>               plugin.store_output("test_output", "/tmp/test_file.txt")

test/test_plugins_core.py:128: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/dirac_cwl_proto/execution_hooks/plugins/core.py:145: in store_output
    self._datamanager.putAndRegister(str(lfn), src, "test")
.pixi/envs/default/lib/python3.12/site-packages/DIRAC/DataManagementSystem/Client/DataManager.py:440: in putAndRegister
    res = self.__hasAccess("addFile", lfn)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.pixi/envs/default/lib/python3.12/site-packages/DIRAC/DataManagementSystem/Client/DataManager.py:119: in __hasAccess
    res = self.fileCatalog.hasAccess(paths, opType)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <DIRAC.Resources.Catalog.FileCatalog.FileCatalog object at 0x7fb88ed76bd0>
name = 'hasAccess'

    def __getattr__(self, name):
        self.call = name
        if name in self.write_methods:
            return self.w_execute
        elif name in self.ro_methods:
            return self.r_execute
        else:
>           raise AttributeError
E           AttributeError

.pixi/envs/default/lib/python3.12/site-packages/DIRAC/Resources/Catalog/FileCatalog.py:158: AttributeError
```

arrabito:
> > Note also that in the current implementation of the get_input_query method, transformation inputs are obtained through a hard-coded rule, specific to a given workflow/use case, so that it’s not very convenient.
> 
> I am not sure to understand why it's not convenient, do you have an example?
> 

For me it's not practical because you have to implement a specific plugin for each different workflow. 
But more important, when we will submit transformations to diracx, I guess that input queries will be treated by a dedicated task (as it's done now in DIRAC legacy by the InputDataQueryAgent), so the fact of implementing the logic of getting input files in a hook plugin does not seem correct to me. When we will submit productions/transformations to diracx we will not specify the get input files logic in the hook plugin or am I missing something?
 
> > Now it’s good to mimic the transformation execution locally, but I’m not sure that we need to keep such complexity, i.e. dynamic jobs creation as soon as input files appear.
> 
> It's mostly used when we launch a production with multiple transformations. I think it's good to show that it's feasible, but I agree that it's not ideal because the current prototype relies on threads, whereas we will depend on `diracx-tasks` in `diracx`. But why would we get rid of it for local tests?
> 

I see, it's true that for  production submission it makes sense to have such a feature. However, I don't really like the fact that the get input files logic is implemented in a hook plugin, because as I said before in diracx it will be handled by a diracx task. Maybe we should think at an alternative to keep this feature but not in the logic of the hook plugin.
Any idea?

> > So I propose that if we want to keep transformation local execution, we could just support either transformations without input files either data-processing transformations with a static list of input files (supposed to be already present on the local host).
> 
> So how do we deal with productions? Does it mean it's not so useful for you?
> 

You mean dealing with productions for local execution?
I think it's useful, I didn't realized that it was needed for productions, but as I said before, I think that we should find an alternative implementation.

> > (I’m going to open a separate issue about the implementation of the DIRAC submission client for transformations).
> 
> Ok great! Do we want to submit CWL transformations to DIRAC? Is that something expected on your side?

I wrote DIRAC to mean diracx.

src/dirac_cwl_proto/execution_hooks/data_management/data_manager.py
mexanick: It is not enough to patch DataManager without specifying any logic, as you need to register a FC first. Below is a minimal working example with an in-memory FC. You can follow it to create a local FS-based FC and base DM on it. 

```python
from DIRAC import S_OK, S_ERROR
from DIRAC.DataManagementSystem.Client.DataManager import DataManager
from DIRAC.Resources.Catalog.FileCatalog import FileCatalog
import os

# ---------------------------
# In-memory FileCatalog
# ---------------------------
class InMemoryFileCatalog(FileCatalog):
    """Minimal in-memory FileCatalog compatible with DIRAC DataManager."""

    def __init__(self, catalogs=None, vo=None):
        self._eligibleCatalogs = {}
        self._files = {}  # store metadata and logical file names
        super(FileCatalog, self).__init__()

    def _getEligibleCatalogs(self):
        self._eligibleCatalogs = {"MyMockCatalog": {"Type": "MockFileCatalog", "Backend": "Memory"}}
        return S_OK(self._eligibleCatalogs)

    def findFile(self, lfn):
        if lfn in self._files:
            return S_OK([self._files[lfn]])
        return S_ERROR(f"File {lfn} not found")

    def addFile(self, lfn, metadata=None):
        if lfn in self._files:
            return S_ERROR(f"File {lfn} already exists")
        self._files[lfn] = {"LFN": lfn, "Metadata": metadata or {}}
        return S_OK(lfn)


# ---------------------------
# Mock DataManager
# ---------------------------
class MockDataManager(DataManager):
    """DataManager mock using an in-memory FileCatalog."""

    def __init__(self, catalogs=None, vo=None):
        self.fileCatalog = InMemoryFileCatalog()
        self._storage = {}  # LFN -> file content

    # Signature similar to DataManager.getFile(lfn, path=None, ... )
    def getFile(self, lfn, path=None, overwrite=False, timeout=-1, useCache=True):
        result = self.fileCatalog.findFile(lfn)
        if not result["OK"]:
            return result
        content = self._storage.get(lfn)
        if content is None:
            return S_ERROR(f"No content stored for {lfn}")
        # Optionally write to path if provided
        if path:
            with open(path, "w") as f:
                f.write(content)
        return S_OK(content)

    # Signature similar to DataManager.putAndRegister(localFile, lfn, metadata=None, ... )
    def putAndRegister(self, localFile, lfn, metadata=None, overwrite=False, timeout=-1, checkProxy=True):
        if not os.path.exists(localFile):
            return S_ERROR(f"Local file {localFile} does not exist")
        with open(localFile, "r") as f:
            content = f.read()
        self._storage[lfn] = content
        return self.fileCatalog.addFile(lfn, metadata=metadata)


# ---------------------------
# Example usage
# ---------------------------
# Write a local dummy file
with open("dummy.txt", "w") as f:
    f.write("Hello in-memory DIRAC")

dm = MockDataManager()

# Put and register
print(dm.putAndRegister("dummy.txt", "/test/file1", metadata={"size": 20}))

# Get file content
print(dm.getFile("/test/file1"))
# Write to a path
print(dm.getFile("/test/file1", path="retrieved.txt"))

# Check catalog
print(dm.fileCatalog._eligibleCatalogs)
print(dm.fileCatalog._files)
```

Expected output:
```shell
python min_data_manager.py
/Users/mdalchen/miniforge3/envs/dirac-test/lib/python3.13/site-packages/DIRAC/__init__.py:63: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  from pkg_resources import get_distribution, DistributionNotFound
{'OK': True, 'Value': '/test/file1'}
{'OK': True, 'Value': 'Hello in-memory DIRAC'}
{'OK': True, 'Value': 'Hello in-memory DIRAC'}
{}
{'/test/file1': {'LFN': '/test/file1', 'Metadata': {'size': 20}}}
```

src/dirac_cwl_proto/execution_hooks/plugins/core.py
```diff
@@ -39,12 +45,108 @@ class QueryBasedPlugin(ExecutionHooksBasePlugin):
...
-    def __init__(self, **data):
-        super().__init__(**data)
-        # Create data catalog with current parameters
-        self.data_catalog = DefaultDataCatalogInterface(
-            campaign=self.campaign,
-            site=self.site,
-            data_type=self.data_type,
-            base_path=self.query_root,
-        )
+    def get_input_query(
```
mexanick: We have previously agreed to remove these methods from `ExecutionHooksPlugin` (and derived classes), as they are parts of the FC. Please do not add them back

Loxeris: These methods are currently present in the `ExecutionHooksPlugin`. I had moved them to `QueryBasedPlugin` but I plan to remove them and use FC methods (`findFileByMetadata` IIRC) in a seperate PR, as it needs changes in the transformation submission as well.

src/dirac_cwl_proto/execution_hooks/core.py
```diff
@@ -240,24 +76,24 @@ class ExecutionHooksBasePlugin(BaseModel):
     version: ClassVar[str] = "1.0.0"
     description: ClassVar[str] = "Base metadata model"
 
-    # Private attribute for data catalog interface - not part of Pydantic model validation
-    _data_catalog: Optional[DataCatalogInterface] = PrivateAttr(
-        default_factory=lambda: DefaultDataCatalogInterface()
+    output_paths: Dict[str, Any] = {}
+    output_sandbox: list[str] = []
+
+    output_se: list[str] = []
```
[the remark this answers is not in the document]
mexanick: I meant DIRAC's code, not our prototype.

src/dirac_cwl_proto/execution_hooks/plugins/core.py
```diff
@@ -39,12 +45,95 @@ class QueryBasedPlugin(ExecutionHooksBasePlugin):
...
+                    res = None
+                    for se in self.output_se:
+                        res = returnSingleResult(
+                            self._datamanager.putAndRegister(str(file_lfn), src, se)
+                        )
+                        if res["OK"]:
+                            break
+                    if res and not res["OK"]:
+                        raise RuntimeError(
+                            f"Could not save file {src} with LFN {str(lfn)} : {res['Message']}"
```
aldbr: Do we want to specify `store_output()` in the `QueryBasedPlugin`?
I haven't checked carefully (so please let me know if I'm wrong), but it seems to be a copy paste of what is defined in `ExecutionHooksHint`.

If it's a duplicate, I would remove it from here.

src/dirac_cwl_proto/job/__init__.py
```diff
@@ -186,15 +194,47 @@ def prepare_input_sandbox(input_data: dict[str, Any]) -> list[Path]:
...
-        file_path = Path(file.path.replace("file://", ""))
-        files_path.append(file_path)
+        if file.path:
+            file_path = Path(file.path.replace("file://", ""))
+            files_path.append(file_path)
 
     return files_path
 
 
+def get_lfns(input_data: dict[str, Any]) -> dict[str, Path | list[Path]]:
```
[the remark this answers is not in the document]
aldbr: Sorry I don't remember: why do we need to separate `lfns_input` exactly? (at some point, we may need to have them separated indeed, to schedule the jobs where the input data is; I am just wondering if we should separate them on the client side)

src/dirac_cwl_proto/transformation/__init__.py
```diff
@@ -138,6 +144,7 @@ def submit_transformation_router(transformation: TransformationSubmissionModel)
         for input_name, group_size in transformation_execution_hooks.group_size.items():
             # Get input query
             logger.info(f"\t- Getting input query for {input_name}...")
+            assert isinstance(transformation_metadata, QueryBasedPlugin)
```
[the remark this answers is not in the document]
aldbr: I would prefer to have `get_input_query` in `ExecutionHooksHintPlugin` rather than in `QueryBasedPlugin` to avoid that (we are experimenting LHCb workflows on our side, and this would prevent us from executing a transformation)

src/dirac_cwl_proto/job/__init__.py
```diff
@@ -62,6 +62,11 @@ async def submit_job_client(
         PrototypeSubmissionClient() if local else DIRACSubmissionClient()
     )
 
+    if local:
+        os.environ["DIRAC_PROTO_LOCAL"] = "1"
+    else:
+        os.environ["DIRAC_PROTO_LOCAL"] = "0"
```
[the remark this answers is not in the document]
Loxeris: It's possible, but the condition is also present in `/transformation/__init__.py` and `production/__init__.py`. These don't make use of `submission_clients.py` so they would need to instantiate it just to set the env variable or keep the condition

src/dirac_cwl_proto/transformation/__init__.py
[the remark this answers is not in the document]
Loxeris: Okay, but I don't think there will be any difference between `ExecutionHooksBasePlugin` and `QueryBasedPlugin` if I move the query methods back.

test/test_workflows.py
```diff
@@ -263,14 +269,56 @@ def test_run_job_parallely():
...
+                    "test/workflows/pi/type_dependencies/job/result_3.sim",
+                    "test/workflows/pi/type_dependencies/job/result_4.sim",
+                    "test/workflows/pi/type_dependencies/job/result_5.sim",
+                ]
+            },
+        ),
+    ],
+)
+def test_run_job_with_input_data(
+    cli_runner, cleanup, pi_test_files, cwl_file, inputs, destination_source_input_data
```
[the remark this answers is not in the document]
aldbr: Why don't we need to prepare the `filecatalog` directory in other test?

src/dirac_cwl_proto/job/__init__.py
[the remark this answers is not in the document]
aldbr: Yes indeed.
Wouldn't it be simpler to:
- not set `DIRAC_PROTO_LOCAL` at all by default
- set it to "1" at the beginning of the `/jobs/__init__.py` `submit_job_router` (because then it means the submission is done locally).

You would avoid having to define it everywhere because:
- if `submit_job_router` is not called, then the submission is done through DIRAC
- if `submit_job_router` is called, then the submission is done locally

Is is clear? :sweat_smile:

test/test_workflows.py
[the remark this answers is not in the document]
Loxeris: It's also done in the transformation tests IIRC.
The usual `test_run_job_success` workflows use local files as inputs, so they are uploaded in a sandbox instead. There's no inputs that are expected to be on the FileCatalog for these workflows.

src/dirac_cwl_proto/transformation/__init__.py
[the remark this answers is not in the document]
aldbr: Ah yes I see.
Well, I guess `ExecutionHooksHintPlugin` is expected to be abstract, unusable per se (well it is not abstract in practice).
And `QueryBasedPlugin` is just 1 concrete and simple example with no pre/post execution commands.

src/dirac_cwl_proto/job/__init__.py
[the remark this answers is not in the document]
aldbr: Good point indeed!

PR #40, opened 2025-10-08
feat: resource requirements validation

### LINKED ISSUE #27: Resource Requirement minmax validation
A job containing a `Workflow` (with or without steps) should have valid `ResourceRequirement`.

- [ ] Return an error if the `ResourceRequirement` minmax is invalid (min > max):
       - should be managed by `cwltool` / `cwl-utils`
- [x] Return an error if there's a `ResourceRequirement` requirements in a `Production` workflow
- [x] Add unit tests with examples

Stellatsuu:
I tried to run "bad" jobs with `cwltool` and it seems that it does not check for conflicts or higher requirements than expected in ResourceRequirement.

So, I've come to this code to validate ResourceRequirements in workflows, related to #27. I tried to make something generic in case we need to implement other Requirements in the future?

However, I'm not entirely sure if this matches with how we want to implement things in the proto, so I'd appreciate any feedback or suggestions.


src/dirac_cwl_proto/execution_hooks/requirement_validator.py
```diff
@@ -0,0 +1,217 @@
+from abc import ABC, abstractmethod
+from typing import ClassVar
+
+from cwl_utils.parser.cwl_v1_2 import CommandLineTool, Workflow, WorkflowStep
+from pydantic import BaseModel, ConfigDict
+
+
+class RequirementValidator(BaseModel, ABC):
```
aldbr: Could you explain the purpose of `RequirementValidator` please?
I am not sure to understand why we would need to have this abstract validator.

Do you foresee a need to validate other CWL requirements?
I have the impression that you are thinking ahead and try to abstract your work. It is valuable but may be too premature in this case:
- you have a single inherited class
- I am not entirely sure we want more than that (but may be you have specific ideas in mind)

src/dirac_cwl_proto/job/__init__.py
```diff
@@ -220,6 +228,13 @@ def submit_job_router(job: JobSubmissionModel) -> bool:
 
     # Validate the jobs
     logger.info("Validating the job(s)...")
+
+    try:
+        ResourceRequirementValidator(cwl_object=job.task).validate_requirements()
+    except RequirementError as ex:
+        logger.exception(f"RequirementValidationError - {ex}")
+        raise
```
aldbr: I am not sure to understand why you catch the error here and then raise it again?

Stellatsuu: If I remember correctly, it wasn't returning anything when calling bad requirements tests (since it's not the client but the server sending the error), and so, the test were failing, since they're waiting for an error from the client that doesn't occur. 

I'll check again this week, since I changed how the test works now

Stellatsuu:
@aldbr Here's an update for now :

I removed my Validator classes and added everything into `SubmissionsModels` as asked : 
- Every type of WF (Job, Transformation, Production) goes through [the same validation before the model is created](https://github.com/aldbr/dirac-cwl-proto/pull/40/files#diff-40bf4fbbe4ac0ed14e73c4c0ec3614f70d9308be6f60eb683efdba13bac4cfefR57-R65). 
      - I don't know if I should've included [the methods](https://github.com/aldbr/dirac-cwl-proto/pull/40/files#diff-40bf4fbbe4ac0ed14e73c4c0ec3614f70d9308be6f60eb683efdba13bac4cfefR160-R244) in the SubmissionModel class or if I let them as module helpers outside of it. 
      - Once we add every possible check in `cwl-utils`/`cwltool`, we'll be able to reduce the validation cases/code and maybe even remove all of it.
      - Should I also add this validation into `TransformationSubmissionModel` and `ProductionSubmissionModel` ? I saw we always end up calling `JobSubmissionModel`, that's why I only implemented it there.
- In case of a Production, there's [a special validation alongside the steps-metadata validation.](https://github.com/aldbr/dirac-cwl-proto/pull/40/files#diff-40bf4fbbe4ac0ed14e73c4c0ec3614f70d9308be6f60eb683efdba13bac4cfefR115-R138)

As for the implementation in `cwl-utils`/`cwltool`, I'm working on it (see: https://github.com/aldbr/dirac-cwl-proto/pull/40#discussion_r2494455977). I think it would be great to add the minmax check in both codes, what do you think? Do you think I should ask the devs?

Stellatsuu:
Open `cwl-utils` PR: https://github.com/common-workflow-language/cwl-utils/pull/384
TODO `cwl-tool` PR: working on it

aldbr (changes requested):

aldbr (changes requested):

aldbr (approved):

natthan-pigoux (approved):

Issue #43, opened 2025-10-20
Use the `--parallel` option


aldbr:
From what I understand, by default `cwltool` executes each step sequentially, even if they are not depending from each other.
I think we want to execute the steps in parallel whenever it's possible in the worker nodes.

Once should:
- [ ] make sure workflows run sequentially without the `--parallel` option
- [ ] add the `--parallel` option 
- [ ] add unit tests (make sure it works within the cpu limits of the worker node, ...)

mr-c:
I don't remember how you are using `cwltool`. Feel free to customize it however you wish. For Python based CWL runners that leverage `cwltool` they typically call into the `cwltool` modules directly instead of using the `cwltool` CLI.


 There is also a `CWLTOOL_OPTIONS` environment variable https://cwltool.readthedocs.io/en/latest/index.html#run-on-the-command-line

PR #48, opened 2025-11-04
Parallel workflow execution

AcquaDiGiorgio:
Add parallel execution at job submission by default

Tests check the following: 
- If the workflow got actually executed parallely (against a sequential workflow)
- Having only a singular available core, if it acts like a sequential workflow

See #43

### DIFF
--- a/test/test_workflows.py
+++ b/test/test_workflows.py
@@ -270,7 +270,7 @@ def test_run_job_validation_failure(
         )
 
 
-def test_run_parallely(cli_runner, cleanup):
+def test_run_job_parallely():
     error_margin_percentage = 0.10
 
     # This command forces the process 'dirac-cwl' to execute ONLY in


aldbr (commented):
Great! Thank you @AcquaDiGiorgio 🙂 

I just have a few comments:
- We are streamlining the tests in parallel and the mandelbrot workflow will likely go away. Do you think you could get the same results using the `crypto` workflow? I think it contains 4 independent steps.
- Do you think we need to have a "sequential" version of the tested workflow? Naively, I think it is not necessary because it's already tested everywhere else: if dependent steps would run in parallel, other tests couldn't work. 
- I think I would test the `crypto` workflow with `taskset -c 0` and then `taskset -c 0-1` for instance to demonstrate the difference in duration.

Do you have any opinion? Please let me know if you think I 'm wrong

AcquaDiGiorgio:
Due to the nature of the crypto algortithms, it might be difficult to differentiate between a parallel and a sequential execution, in both cases they get executed extremely quick. If `pi-simulate` is going to stay in the future, I might be able to use it, because the size of inputs will vary the execution time.
 
I'm using a forced sequential workflow to be able to compare it with a parallel one that executes in one core. 
If you find this test innecesary, I can completely remove it and the sequential cwl, bacause for first test, as you said, the sequential workflow can be a `taskset -c 0` of the parallel one.

Issue #16, opened 2025-08-06
Command pattern (pre/post processing steps)


arrabito:
About dirac-metadata schema, in:

   
[     enum: ["User", "Admin", "Guest"]  # Example enum values](https://github.com/aldbr/dirac-cwl-proto/blob/main/src/dirac_cwl_proto/schemas/dirac-metadata.yaml#L10)

shouldn't we set as allowed values only the Models defined here:

https://github.com/aldbr/dirac-cwl-proto/blob/main/src/dirac_cwl_proto/metadata_models.py

?

i.e.:

`enum: ["User", "TaskWithMetadataQuery", "PiSimulate", ....] 
`
etc?

Moreover did I understand correctly that now the idea is to keep only 2 generic models (or a few additional ones if needed), like 'User' and 'TaskWithMetadataQuery', instead of creating a new one for each type of job?

If my understanding is correct, then we should just keep:

`enum: ["User", "TaskWithMetadataQuery"]`

Is that correct?

arrabito:
> I am not entirely sure about that, I am still trying to figure out how it would work for LHCb.
Right now in LHCb we have different post-processing steps after the workflow execution, depending on the job type (e.g. MCSimulation, User, Merge, ...)

I see. In CTAO for now we don't have different post-processing steps depending on the job type, but eventually we could also have this use case.

In current DIRAC and extensions this is handled by implementing different Workflow Modules:
https://github.com/DIRACGrid/DIRAC/tree/integration/src/DIRAC/Workflow/Modules
Is that correct?

Anyway, I haven't thought yet about how to handle different post-processing, but I guess that this could be done in the extensions, while in diracx we could provide a generic post-processing to handle the output data.

Does it make sense?

aldbr:
> In current DIRAC and extensions this is handled by implementing different Workflow Modules:
https://github.com/DIRACGrid/DIRAC/tree/integration/src/DIRAC/Workflow/Modules
Is that correct?

Yes it is correct.

> Anyway, I haven't thought yet about how to handle different post-processing, but I guess that this could be done in the extensions, while in diracx we could provide a generic post-processing to handle the output data.

This is indeed the direction we are taking.
Right now extensions can create their own `execution hooks` per "job type" (`execution hooks plugin`).

But they would potentially have to redefine the logic by themselves: if they want to change a minor part of the base execution hooks without to change the others, we need to provide them with convenient tools to do so, without having to rewrite everything.

[...]

PR #55, opened 2025-11-19
feat: Add hints to pre-process and post-process jobs depending on its job type

AcquaDiGiorgio:
This is my approach to solve issue #16 

I'm sure this solution is too complex and does not follow the expected structure and that's why I open this PR as a draft. I whould kindly ask for some insights.
It's also a draft because it is not a completely robust solution. There are a couple of edge cases that are not covered yet, such as if a JobType does not exist.

My issue in terms "unknowns" is that I'm not sure where the JobType should reside.
- Should there be a Plugin for each type of Job? 
- If so, could they somehow be combined with the QueryBasedPlugin (If this makes any sense) or any non "JobType" Plugin?

### DIFF
--- a/src/dirac_cwl_proto/job/job_wrapper.py
+++ b/src/dirac_cwl_proto/job/job_wrapper.py
@@ -192,7 +192,6 @@ class JobWrapper:
 
             # Post-process the job
             logger.info("Post-processing Task...")
-
             if self._post_process(
                 result.returncode,
                 result.stdout,
--- a/test/test_execution_hooks_core.py
+++ b/test/test_execution_hooks_core.py
@@ -5,6 +5,7 @@ This module tests the foundational classes and interfaces of the execution hooks
 plugin system, including ExecutionHooksBasePlugin, ExecutionHooksHint, and core
 abstract interfaces.
 """
+
 from pathlib import Path
 from typing import Any, List, Optional, Union
 


src/dirac_cwl_proto/execution_hooks/core.py
```diff
@@ -373,6 +376,40 @@ def from_cwl(cls: type[T], cwl_object: Any) -> T:
...
+
+    model_config = ConfigDict(extra="forbid", validate_assignment=True)
+
+    jobType: Optional[str] = Field(default=None, description="")
+
+    def instantiate(self, jobWrapper) -> JobProcessorBase:
+        if self.jobType is None:
+            return JobProcessorBase(jobWrapper=None)
+
+        if not isinstance(self.jobType, str) or "/" not in self.jobType:
```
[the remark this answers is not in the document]
AcquaDiGiorgio: I thought that too, in relation to using entrypoints, but as I am not that experienced with them I tried with this approach instead.
I will look into this and make some modifications 👌

src/dirac_cwl_proto/commands/generic/jobTypes.py
```diff
@@ -0,0 +1,8 @@
+from dirac_cwl_proto.commands import JobProcessorBase
+
+from .modules import DownloadConfig, GroupOutputs
+
+
+class TestingJobType(JobProcessorBase):
+    preprocess_commands = [DownloadConfig]
+    postprocess_commands = [GroupOutputs]
```
aldbr: I think this is the right approach, but ideally we should move these into https://github.com/aldbr/dirac-cwl-proto/blob/5ed828cc018d87151e407f3c4fa43d62f18cf206/src/dirac_cwl_proto/execution_hooks/plugins/core.py#L19

src/dirac_cwl_proto/commands/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+
+class CommandBase(ABC):
+    @abstractmethod
+    def execute(self, job_wrapper):
+        raise NotImplementedError()
+
+
+class JobProcessorBase:
+    preprocess_commands: List[type[CommandBase]] = []
+    postprocess_commands: List[type[CommandBase]] = []
```
aldbr: And this should be moved into https://github.com/aldbr/dirac-cwl-proto/blob/5ed828cc018d87151e407f3c4fa43d62f18cf206/src/dirac_cwl_proto/execution_hooks/core.py#L218

Note: you can see that `pre_process` and `post_process` already exist, and you can just add your logic there.

src/dirac_cwl_proto/execution_hooks/__init__.py
```diff
@@ -11,6 +11,7 @@
     DataCatalogInterface,
     ExecutionHooksBasePlugin,
     ExecutionHooksHint,
+    PrePostProcessingHint,
```
aldbr:
```suggestion
```

src/dirac_cwl_proto/job/job_wrapper.py
```diff
@@ -170,8 +170,13 @@ def run_job(self, job: JobSubmissionModel) -> bool:
         self.job_path.mkdir(parents=True, exist_ok=True)
 
         try:
+            jobProcessor = job.jobProcessing.instantiate(self)
```
aldbr: Use `self.runtime_metadata` instead (may be this should be renamed as `self.execution_hooks`) 
```suggestion
```

src/dirac_cwl_proto/job/job_wrapper.py
```diff
@@ -193,6 +198,10 @@ def run_job(self, job: JobSubmissionModel) -> bool:
 
             # Post-process the job
             logger.info("Post-processing Task...")
+
+            # If we add to JobWrapper class a 'result' attribute, it could be accessed
+            #  by the jobProcessor and process this with a single post_process function
+            jobProcessor.post_process()
```
aldbr:
```suggestion
```

src/dirac_cwl_proto/submission_models.py
```diff
@@ -133,11 +138,17 @@ def serialize_task(self, value):
 # -----------------------------------------------------------------------------
 
 
-def extract_dirac_hints(cwl: Any) -> tuple[ExecutionHooksHint, SchedulingHint]:
-    """Thin wrapper that returns (ExecutionHooksHint, SchedulingHint).
+def extract_dirac_hints(
+    cwl: Any,
+) -> tuple[ExecutionHooksHint, SchedulingHint, PrePostProcessingHint]:
```
aldbr:
```suggestion
) -> tuple[ExecutionHooksHint, SchedulingHint]:
```

src/dirac_cwl_proto/submission_models.py
```diff
@@ -133,11 +138,17 @@ def serialize_task(self, value):
...
+    """Thin wrapper that returns (ExecutionHooksHint, SchedulingHint, PrePostProcessingHint).
```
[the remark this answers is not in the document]
aldbr: Oops I think you forgot this comment

aldbr (commented):
Great @AcquaDiGiorgio!
You are on the right track, it is much more simpler.
I think we can even make it simpler based on my comments, but I am open to your suggestions if you think the `JobTypeProcessorBase` is needed.

src/dirac_cwl_proto/commands/download_config.py
```diff
@@ -0,0 +1,18 @@
+import os
+
+from dirac_cwl_proto.commands import CommandBase
+
+
+class DownloadConfig(CommandBase):
```
ryuwd: Any reason that this can't just be a function ? I don't see much benefit of using abstract classes

@aldbr what do you think?

src/dirac_cwl_proto/commands/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+    @classmethod
+    def pre_process(cls, job_path: Path, **kwargs) -> None:
+        for command in cls.preprocess_commands:
+            try:
+                command().execute(job_path, **kwargs)
+            except Exception as e:
+                logger.error(
+                    f"Command '{command.__name__}' failed during the pre-process stage: {e}"
+                )
+                raise
```
ryuwd:
```suggestion
            try:
                command().execute(job_path, **kwargs)
            except Exception as e:
                logger.exception(
                    f"Command '{command.__name__}' failed during the pre-process stage"
                )
                raise WorkflowExecutionExceptionOrSomeOtherName("Command '{command.__name__}' failed during the pre-process stage") from e
```

Unsure if the new exception is necessary but maybe be a nice-to-have

src/dirac_cwl_proto/commands/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+    @classmethod
+    def post_process(cls, job_path: Path, **kwargs) -> None:
+        for command in cls.postprocess_commands:
+            try:
+                command().execute(job_path, **kwargs)
+            except Exception as e:
+                logger.error(
+                    f"Command '{command.__name__}' failed during the post-process stage: {e}"
+                )
+                raise
```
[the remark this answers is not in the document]
aldbr: > We can move the way of discovering plugins to the entrypoints, but that's another problem.
Indeed, and it's up to you. Either you add the entrypoints mechanism in this PR, or we create an issue for that.

> The discovery is not an issue if the plugin can be treated as the job type, because in such case, JobTypeProcessorBase would be removed and nothing would need to be discovered.

The plugin, so far, IS a job type. We just named it differently for now.

src/dirac_cwl_proto/commands/__init__.py
```diff
@@ -0,0 +1,56 @@
+import logging
+from abc import ABC, abstractmethod
+from pathlib import Path
+from typing import List
+
+logger = logging.getLogger(__name__)
+
+
+class CommandBase(ABC):
```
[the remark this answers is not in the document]
AcquaDiGiorgio: I like this proposal, but I think having two arrays is the better option.

If we only use one, a plugin could have the following configuration: 
`commands=[PreProc1, PreProc2, PostProc1, PreProc3, PostProc2]`

Also, if the commands do not have "Pre" or "Post" in their name, it might be difficult to know when the preprocess ends and postprocess starts. Imagine the following config:
```python
commands = [
 CreateDataFile,
 FileUsage,
 AnalyseXMLSummary,
 BookkeepingReport,
 StepAccounting,
 UploadOutputData,
 RemoveInputData,
 UploadLogFile,
 UploadMC,
 FailoverRequest,
]
```
is much harder to visualize than:
```python
preprocess_commands = [
 CreateDataFile,
 FileUsage,
]
postprocess_commands = [
 AnalyseXMLSummary,
 BookkeepingReport,
 StepAccounting,
 UploadOutputData,
 RemoveInputData,
 UploadLogFile,
 UploadMC,
 FailoverRequest,
]
```
> so far, I don't think we have any use case where we have 1 command that could be used both as a pre and post processing command

If we use 2 arrays and somewhere in the future someone decides to create a command with this requirement, it would be as easy as inherit both classes and use it in both arrays.

aldbr (commented):
LGTM, you would just need to address @ryuwd's comment about the exceptions: https://github.com/aldbr/dirac-cwl-proto/pull/55#discussion_r2559607190

src/dirac_cwl_proto/commands/core.py
```diff
@@ -0,0 +1,27 @@
+from abc import ABC, abstractmethod
+from pathlib import Path
+
+
+class CommandBase(ABC):
```
ryuwd:
```suggestion
class _CommandBase(ABC):
```

src/dirac_cwl_proto/commands/core.py
```diff
@@ -0,0 +1,27 @@
...
+        raise NotImplementedError("This method should be implemented by child class")
+
+
+class PreProcessCommand(CommandBase):
+    """Interface class for pre-processing commands.
+    Every pre-processing command must inherit this class. Used for type validation.
+    """
+
+
+class PostProcessCommand(CommandBase):
```
ryuwd:
```suggestion
class PostProcessCommand(_CommandBase):
```

test/test_plugins_core.py
```diff
@@ -192,3 +196,130 @@ def test_plugin_serialization_compatibility(self):
             schema = plugin.model_json_schema()
             assert isinstance(schema, dict)
             assert "properties" in schema
+
+
+class TestPluginCommands:
```
aldbr: One very last personal suggestion about the tests in general (we can discuss about them, I am open to your feedback if you have a different opinion).

I think tests can be very interesting, for newcomers, to better understand how a given component works, or should work.
I think we should, in general, try to make them very clear, even if it implies some duplication of code.

- In general I avoid parameterize them because it's generally clearer to read something like:
  ```python
  with pytest.raises(WorkflowProcessingException):
  ```
  rather than:
  ```
  if exception:
      with pytest.raises(WorkflowProcessingException):
  else: ...
  ```

- Making fixture in this case can make sense indeed, but I think the name should try to provide as much info as needed for the reader to understand without having to read the fixture code. If the reader has to go through the fixture code, then I would add further comments to guide the reader.

Note: making good unit test is very complex task 😅

Issue #62, opened 2025-11-27
feat: Implement a "JSON catalog" input data resolution format


ryuwd:
Relates to https://github.com/aldbr/dirac-cwl-proto/issues/25#issuecomment-3481250621, https://github.com/DIRACGrid/DIRAC/issues/8309

- [x] Create a pydantic model

`Done` criteria:
- the model should follow specification described in the DIRAC issue

cc @aldbr

ryuwd:
See https://github.com/DIRACGrid/diracx/pull/741

PR #63, opened 2025-11-27
feat: pydantic model for json catalog and a generator w/ input data resolution

ryuwd:
for https://github.com/aldbr/dirac-cwl-proto/issues/62

### DIFF
--- a/src/dirac_cwl_proto/job/replica_catalog.py
+++ b/src/dirac_cwl_proto/job/replica_catalog.py
@@ -1,3 +1,5 @@
+import enum
+import logging
 import re
 from typing import Annotated
 
@@ -9,6 +11,11 @@ from pydantic import (
     RootModel,
     field_validator,
 )
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+
+from DIRAC.Core.Utilities.ReturnValues import returnValueOrRaise
 
 
 def validateLFN(value: str) -> str:
@@ -101,7 +108,7 @@ class ReplicaCatalog(RootModel):
                 None  # Like 6032CB7C-32DC-EC11-9A66-D85ED3091D71
             )
 
-            @field_validator("adler32", "md5", "guid")
+            @field_validator("adler32", "guid")
             @classmethod
             def validate_at_least_one_checksum(cls, v, info):
                 # This validator runs for each field, so we check after all fields are processed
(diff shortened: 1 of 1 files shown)


src/dirac_cwl_proto/job/json_catalog.py
```diff
@@ -0,0 +1,37 @@
...
+
+# Logical File Name such as LFN:/lhcb/MC/2024/HLT2.DST/00327923/0000/00327923_00000533_1.hlt2.dst
+LFN = Annotated[str, BeforeValidator(validateLFN)]
+PFN = Annotated[AnyUrl, BeforeValidator(validatePFN)]
+StorageElementId = str
+ChecksumType = str
+FileChecksum = str
+
+
+class JSONCatalog(RootModel):
```
aldbr: I'm just wondering if it's the best name we can come up.
What about `ReplicaCatalog`?

src/dirac_cwl_proto/job/json_catalog.py
```diff
@@ -0,0 +1,37 @@
...
+    class CatalogEntry(BaseModel):
+        class Replica(BaseModel):
+            url: PFN
+            se: StorageElementId
+        replicas: list[Replica]
+        size_bytes: int | None = None
+        checksum: dict[ChecksumType, FileChecksum] | None = None
```
aldbr: I'm just wondering if we need to support multiple checksums.
If yes, may be we could validate the `FileChecksum` based on the `ChecksumType`.

aldbr: Well actually this JSON object will be generated programmatically and I don't see any use case where a user would manually create it. Therefore, I am not sure it actually makes much sense to make tests only around the pydantic model.
I think what we want to have and test in this PR is the code generating this JSON object from a CWL input file.

Example:
- I have this input:

```
input-data:
  - class: File
    path: lfn://path/to/result_1.sim
```

And I expect to get the following output from it:

```json
{
  ...
}
```

Any opinion?

ryuwd: I'm inclined to support multiple checksum types particularly to support e.g. Dirac/Gaudi File GUIDs, which facilitates converting a replica catalog to pool XML catalog format

I absolutely agree we should validate checksum based on the Checksum type, in which case the checksum field should be a model

src/dirac_cwl_proto/job/replica_catalog.py
```diff
@@ -0,0 +1,154 @@
...
+            @classmethod
+            def validate_se(cls, v: str) -> str:
+                if not v or not v.strip():
+                    raise ValueError("Storage Element ID cannot be empty")
+                return v.strip()
+
+        class Checksum(BaseModel):
+            crc32: str | None = None  # ???
+            adler32: Adler32Checksum | None = None  # like 788c5caa
+            md5: MD5Checksum | None = None  # like d41d8cd98f00b204e9800998ecf8427e
```
ryuwd:
```suggestion
```

PR #64, opened 2025-11-27
feat(ExecutionHooksPluginRegistry): Move ExecutionHooksPlugin discovery to pyproject entrypoints

AcquaDiGiorgio:
Changed the way of discovering plugins to utilize pyproject entrypoints. 
The name of the entrypoint is unused.

## test/test_exectuion_hooks_registry.py
`test_discover_plugins` renamed to `test_discover_no_plugins` and adapted to the new code.
`test_discover_plugins` now checks that the plugin is the correct subclass and if it is not, that it gets logged correctly.


aldbr (approved):
LGTM 🙂

Issue #73, opened 2025-12-12
Use diracx-api jobs.py module to interact with the sandbox store endpoint (instead of the one from DIRAC)


aldbr:
Here we have a class that mimic the DIRAC `SandboxStoreClient`: 

- mock: https://github.com/aldbr/dirac-cwl-proto/blob/a1900b590012ae03541dd90267298d4e5ba3c6d0/src/dirac_cwl_proto/data_management_mocks/sandbox.py
- real: https://github.com/DIRACGrid/DIRAC/blob/bdcf074c302c78ec06101864b7eee27a50015a36/src/DIRAC/WorkloadManagementSystem/Client/SandboxStoreClient.py

The `sandbox` management is already supported within `diracX` and should be preferred:
- https://github.com/DIRACGrid/diracx/blob/321e409fc3b0d24fd6534e38a47063b95407dabf/diracx-api/src/diracx/api/jobs.py

**Tasks**
- [ ] Replace the `sandbox_store_client` by calls `diracx-api.jobs`
- [ ] Replace the mock `sandbox_store_client` for local execution by a fake `diracx-api.jobs` module
- [ ] If anything is missing in `diracx`, let's first discuss and see if we should try to integrate it to `diracx`

**Done criteria**
- [ ] existing tests should still pass

PR #89, opened 2026-01-22
feat: sandbox store using diracx api mock

Stellatsuu:
Closes: #73 

Changes:
- added `create_sandbox` and `download_sandbox` functions, mocking `diracx-api` sandboxes management
- added `pytest-asyncio` to create async tests + added a `create_sandbox` test
- removed DIRAC `SandboxStoreClient` and `MockSandboxStoreClient`

### DIFF
--- a/src/dirac_cwl_proto/data_management_mocks/sandbox.py
+++ b/src/dirac_cwl_proto/data_management_mocks/sandbox.py
@@ -53,8 +53,10 @@ def create_sandbox(paths: Sequence[str | Path]):
 
         # Create PFN
         pfn = f"{sandbox_info.checksum_algorithm}:{sandbox_info.checksum}.{sandbox_info.format}"
+        logger.debug("Sandbox PFN is %s", pfn)
 
         # Create sandbox in sandboxstore
+        Path("sandboxstore").mkdir(exist_ok=True)
         sandbox_path = Path(f"sandboxstore/{pfn}")
         if not sandbox_path.exists():
             with tarfile.open(sandbox_path, "w:gz") as tar:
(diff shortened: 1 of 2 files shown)


src/dirac_cwl_proto/data_management_mocks/sandbox.py
```diff
@@ -1,17 +1,103 @@
...
 logger = logging.getLogger(__name__)
 
 
+class MockSandboxInfo(BaseModel):
+    """Mock Dirac Sandbox Info."""
+
+    checksum_algorithm: str
+    checksum: str = Field(pattern=r"^[0-9a-fA-F]{64}$")
+    size: int = Field(ge=1)
+    format: str
```
aldbr: I think you could safely use `SandboxInfo` from `diracx-core` without mocking it 🙂 

```suggestion
from diracx.core import models
```

Unless I miss a blocking point?

src/dirac_cwl_proto/job/job_wrapper.py
```diff
@@ -46,16 +44,10 @@
...
     """Job Wrapper for the execution hook."""
 
-    _sandbox_store_client: SandboxStoreClient
-
     def __init__(self) -> None:
         """Initialize the job wrapper."""
         self.execution_hooks_plugin: ExecutionHooksBasePlugin | None = None
         self.job_path: Path = Path()
-        if os.getenv("DIRAC_PROTO_LOCAL") == "1":
-            self._sandbox_store_client = MockSandboxStoreClient()
```
aldbr: Here we still need the environment variable.
Basically: 
- if we are in a local environment: we want to use the `mock.sandbox`
- if we are in DIRAC: we want to use the `diracx.api.sandbox` API

Do you see what I mean?

PR #93, opened 2026-01-26
refactor: move pre/post_process methods

### LINKED ISSUE #72: Remove `pre_process` and `post_process` from the `ExecutionHooksHint` class (move content to `JobWrapper`)
Now that `ExecutionHooksHint` child classes embed a list of `Pre` and `PostExecution` commands, I think we should move the `pre_process` and `post_process` methods to the `JobWrapper`.

It would better separate responsibilities:
- the `ExecutionHooksHint` child classes would focus on declaring the commands they are going to run
- the `JobWrapper` would handle executing them

arrabito:
Closes #72

### DIFF
--- a/test/test_job_wrapper.py
+++ b/test/test_job_wrapper.py
@@ -105,20 +105,14 @@ class TestJobWrapper:
         """
         from dirac_cwl_proto.commands import PostProcessCommand, PreProcessCommand
 
-        # Initialization
+        # Initialization and set the execute function to raise an exception
         class Command(PreProcessCommand, PostProcessCommand):
             def execute(job_path, **kwargs):
-                return
+                raise NotImplementedError()
 
         plugin = job_type_testing()
         job_wrapper = JobWrapper()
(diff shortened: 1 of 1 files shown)


test/test_job_wrapper.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def test_instantiation(self):
+        """Test that ExecutionHooksBasePlugin can be instantiated directly with default behavior."""
+        hook = ExecutionHooksBasePlugin()
+        job_wrapper = JobWrapper()
+        job_wrapper.execution_hooks_plugin = hook
+
+        # Test default pre_process behavior
+        command = ["echo", "hello"]
+        result = job_wrapper._JobWrapper__pre_process_hooks({}, None, Path("/tmp"), command)
```
aldbr: Same comment here, I think we should use `job_wrapper.pre_process()` and `job_wrapper.post_process()`. What do you think?

Issue #61, opened 2025-11-24
About transformation submission interface


arrabito:
I would like to discuss here some general aspects of the transformation submission interface.

In order to summarize the different use cases for transformations we have :
1. transformations with no input files (let’s call them simulation transformations for simplicity)
2. data-processing transformations, i.e. transformations with input files, where input files are specified as a static list
3. data-processing transformations, i.e. transformations with input files, where input files are specified through an input query

I think that for transformation submission, similarly to what done for job submission, we could implement 2 submission client classes, i.e. one for local submission and one for diracx submission.

**Concerning 1. simulation transformations**, If I’m not wrong in the current prototype we only partially support them. 

It seems to me that we can submit a simulation transformation, but this will generate only 1 job and we cannot specify in the submission interface a varying parameter that will be used to generate several identical jobs. 

Is that correct? 

If yes, we should find a way how to specify this varying parameter for simulation transformations. Any idea?

For reference in legacy DIRAC, the user can set as varying parameter of a ‘simulation’ transformation, a variable dynamically built at run-time from the Task_ID (where ) of each job (i.e. Task_ID = 00001, 00002, …. 0000n_tasks).

**Concerning 2. data-processing  transformations specified through a static list of input files**, I think that we should distinguish 2 use cases, depending if the submission is local or to diracx.

In both cases we could specify the input files in a dedicated field of the `TransformationExecutionHooksHint`, i.e. something like :

```
input-data:   
- class: File     
   path: <path_to_local_file_1 or LFN_1>   
- class: File
   path: <path_to_local_file_2 or LFN_2>
```

where the path is either the path to a local file either a LFN, depending if the submission is local or to diracx.

**Concerning 3. data-processing transformations specified through an input query**

We had initially discussed this topic here: https://github.com/users/aldbr/projects/1/views/1?pane=issue&itemId=130044576&issue=aldbr%7Cdirac-cwl-proto%7C5

but afterwards we have revised several things here: https://github.com/aldbr/dirac-cwl-proto/issues/20

and I’m not sure about how we decided to handle input data queries in the end. 

It seems to me that the idea was to use the `configuration` field of the `ExecutionHooksHint`, which in principle can contain not only metadata key-values for input queries, but any additional parameter used by the hook plugin. Is that correct?

If yes, we could use this field for metadata key-values, e.g.:

```
configuration:   
  <meta_key_1>: <meta_data_1>   
  <meta_key_2>: <meta_data_2>
```

[...]

PR #95, opened 2026-02-03
feat: automatic job grouping

### LINKED ISSUE #66: Support data-processing transformations specified through a static list of input files
## Story

**As a** CTAO transformation manager submitting CWL workflows.

**I want to** specify input files directly in CWL inputs and have them automatically grouped into jobs based on a group size parameter

**So that** I can efficiently process large datasets by distributing files across multiple jobs without manually creating job definitions, and easily keep track of them.

### 1. Enhanced TransformationSubmissionModel
- [x] The `TransformationSubmissionModel` is extended to accept CWL inputs in addition to the CWL workflow
- [x] CWL inputs can contain a list of file paths (either LFNs or local paths)
- [x] The model validates that CWL inputs conform to the expected format

### 2. Input File Specification
- [x] Users can specify a list of input files in the CWL inputs
- [x] Each file path should be LFNs (no sandbox is planned for now)
- [x] The system accepts arbitrarily large lists of input files

### 3. Automatic Job Grouping
- [x] A `group_size` parameter controls how many files are processed per job
- [x] The system automatically calculates the number of jobs based on: `total_files / group_size`
- [x] Files are distributed evenly across jobs according to the group size
- [x] Each job receives its subset of input files from the original list

### 4. Job Creation Logic
- [x] Given N input files and a group size of M, the system creates `N/M` jobs (rounded up if needed)
- [x] Each job is configured to process exactly M files (or fewer for the last job if N is not divisible by M)

### Transformation Submission with 150 Files
```yaml
# CWL Workflow (existing)
workflow: <cwl_workflow_definition>

# CWL Inputs (new)
inputs:
  input_files:
    - lfn://dirac/prod/dataset/file001.root
    - lfn://dirac/prod/dataset/file002.root
    - lfn://dirac/prod/dataset/file003.root
    # ... (147 more files)

# Transformation Hints
group_size: 5
```

**Result**: 
- Total input files: 150
- Group size: 5
- Jobs created: 150 ÷ 5 = **30 jobs**
- Each job processes 5 files

### Job Distribution Example
- **Job 1**: files 1-5
- **Job 2**: files 6-10
- **Job 3**: files 11-15
- ...
- **Job 30**: files 146-150

## Context

This solution reuses the existing CWL inputs mechanism rather than introducing a new option in `TransformationExecutionHooksHint`. The approach provides a clean separation between workflow definition and input data while enabling efficient batch processing of large datasets through automatic job grouping.

Related to the discussion in https://github.com/aldbr/dirac-cwl-proto/issues/61

## Dependencies
- Requires enhancement to `TransformationSubmissionModel` to accept CWL inputs
- Depends on implementation of job grouping logic based on `group_size` parameter

## Technical Notes
- The `group_size` parameter determines the granularity of parallelization
- Users can optimize job submission based on file size and processing requirements
- The solution maintains consistency with CWL standards while adding DIRAC-specific batch processing capabilities.

Stellatsuu:
cc @aldbr @arrabito @natthan-pigoux

Closes: #66 
Related to: #61 

Changes:
- added `input_data: list[str | File]` to `TransformationSubmissionModel` and `ProductionSubmissionModel`
- added optional `inputs-file` parameter to Transformation and Production CLIs: 
`dirac-cwl transformation/production submit file.cwl --inputs-file file.yaml`
- renamed `parameter-path` to `input_files: list[str]` in Job CLI:
`dirac-cwl job submit file.cwl --input-files file1.yaml file2.yaml ...`
- added `group_size` `executionHooksHint` to Transformation Workflows, such as:
```
hints:
  - class: dirac:ExecutionHooks
    group_size: (int)
```
- `group_size` determines the number of jobs to be created and how many inputs files they will contain in `submit_transformation_router`, by default, it equals 1, which mean a job will be created for each input in the inputs file. Once the list of jobs is created, it is sent to the `job_router` and processed.
- added simple tests and workflows (e.g: count and list the inputs files contained in each created job)
- quick fix about `JobWrapper` related tests: `task.cwl` was created during `post_process` but never cleared after running tests.
    
TODO after this PR:
- plan an issue/PR for input sandbox within Transformations if not already done (e.g: https://github.com/DIRACGrid/dirac-cwl/pull/95#discussion_r2773736019)


test/workflows/automatic_job_grouping/inputs_files-strings.yaml
aldbr: We can potentially just try with 15 files with a group size of 3 to start with.
What do you think?
Or even 6 files with a group size of 2 for instance

aldbr: Could we have an example with LFNs too?

src/dirac_cwl_proto/submission_models.py
```diff
@@ -95,15 +95,14 @@ class JobModel(BaseJobModel):
...
-
-
 class TransformationSubmissionModel(BaseModel):
     """Transformation definition sent to the router."""
 
     # Allow arbitrary types to be passed to the model
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     task: CommandLineTool | Workflow | ExpressionTool
+    input_data: Optional[list[str | File] | None] = None
```
aldbr: As we are going to integrate input sandbox within transformations (https://github.com/DIRACGrid/dirac-cwl/issues/92), it would be interesting to see if we could reuse the `JobInputModel` (renamed as `InputModel`?)

src/dirac_cwl_proto/transformation/__init__.py
```diff
@@ -103,39 +110,57 @@ def submit_transformation_router(transformation: TransformationSubmissionModel)
...
-            # Update the input data in the metadata
-            # Only keep the first min_length inputs
-            input_data_dict[input_name] = inputs[:min_length]
-
-        # Get the JobModelParameter for each input
-        job_model_params = _generate_job_model_parameter(input_data_dict)
-        logger.info("Input data for the transformation retrieved!")
+    if transformation.input_data:
+        nb_files = len(transformation.input_data)
+        group_size = (transformation_execution_hooks.group_size or {}).get("input_data") or nb_files
```
aldbr: Ok I see the confusion here and it's my fault.
I originally made `group_size` as a dictionary where a given transformation would have needed X outputs of a given parameter, and Y outputs of another parameter to start a job.

In practice, this is impossible to test locally, and we don't have any use case for that. So this should likely be simplified as `group_size: 5` for instance (need 5 inputs of each parameter before starting a job)

Any opinion @arrabito ?

aldbr (changes requested):

src/dirac_cwl_proto/transformation/__init__.py
```diff
@@ -37,6 +37,7 @@
 @app.command("submit")
 def submit_transformation_client(
     task_path: str = typer.Argument(..., help="Path to the CWL file"),
+    inputs_file: str | None = typer.Option(None, help="Path to the CWL inputs file"),
```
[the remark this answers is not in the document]
arrabito: > For consistency, could you please rename `parameter_path` (and adjust the `help` message) into `input_files` in `job/__init__`?
> 
> I would have been tempted to turn `inputs_file` into a `list[str]` for consistency and scalability but I don't see any use case where one would create 2 similar transformations with different set of input files. @arrabito do you confirm?

@aldbr you mean that if you pass a list of input files, let's say 2 input files, it would result in the creation of 2 transformations, one for each input file? 

If yes, I agree that I don't see any use case for that and I would keep `inputs_file` a str.

src/dirac_cwl_proto/transformation/__init__.py
aldbr: > Is that correct?

Yes it's correct, unless you think it would respond to a use case on your side?

> Is it done differently in this prototype? Does t4 expect 5 files from t2 grouped together and 5 files by t3 grouped together, but not mixing files from t2 and t3 in the same job?

Yes right now it would get 5 files from `t2` and 5 from `t3`. We don't really have such workflows for now in LHCb so I'm happy to change it in a different issue/PR to follow your use case if it's better for you.

> Anyway, this last point I think it's more related to the Production submission, than to this PR. Am I correct?

Yes exactly.

test/test_integration.py
```diff
@@ -95,6 +95,9 @@ def test_user_workflow_scenario(self, sample_job):
         result = job_wrapper.post_process(0, "{}", "{}")
         assert isinstance(result, bool)
 
+        task_file = job_wrapper.job_path / "task.cwl"
+        task_file.unlink(missing_ok=True)
```
aldbr: I see that you are using these 2 lines in different test.
Can you create or reuse an existing fixture within `conftest`?
You can generally `yield` and then add these 2 lines and they will be executed after the execution of the test. Example: https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended

aldbr (changes requested):

natthan-pigoux (changes requested):

natthan-pigoux (changes requested):

natthan-pigoux (changes requested):

natthan-pigoux (dismissed):

src/dirac_cwl/transformation/__init__.py
```diff
@@ -9,8 +9,9 @@
 
 import typer
 from cwl_utils.pack import pack
-from cwl_utils.parser import load_document
+from cwl_utils.parser import load_document, save
 from cwl_utils.parser.cwl_v1_2 import File
```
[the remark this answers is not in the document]
Stellatsuu: `cwl_utils.parser import File` won't work because `File = Union[cwl_v1_0.File, cwl_v1_1.File, cwl_v1_2.File]` in the parser, it's the same issue when you import `CommandLineTool`, `Workflow`, etc., that's why they're imported from `cwl_v1_2`:
```
from cwl_utils.parser import load_document
from cwl_utils.parser.cwl_v1_2 import (
    CommandLineTool,
    ExpressionTool,
    Workflow,
    WorkflowInputParameter,
    WorkflowStep,
)
```
If you try to create a File object with this import, it will fail (`TypeError: 'typing.Union' object is not callable`). Maybe there's another way to do so?

test/test_workflows.py
```diff
@@ -373,6 +369,8 @@ def test_run_nonblocking_transformation_success(cli_runner, cleanup, cwl_file):
     assert "Transformation done" in clean_output, f"Failed to run the transformation: {result.stdout}"
 
 
+# TODO: re-enable this test once approval on questions of new Job Grouping
+@pytest.mark.skip(reason="Waiting for approval on new Job Grouping (the tested code is currently commented)")
```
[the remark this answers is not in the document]
Stellatsuu: Test isn't working anymore so I'll take a look at the code later since it was modified a bit

src/dirac_cwl/transformation/__init__.py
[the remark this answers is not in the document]
natthan-pigoux: That is true only if you create an File() object, but that is not the case here. Anyway this line is actually not part of your changes so I should not have review it.

src/dirac_cwl/submission_models.py
```diff
@@ -118,6 +117,17 @@ def serialize_task(self, value):
         else:
             raise TypeError(f"Cannot serialize type {type(value)}")
 
+    @field_serializer("input_data")
+    def serialize_input_data(self, value):
```
natthan-pigoux: I see that the `TransformationSubmissionModel` and the `ProductionSubmissionModel` share methods and arguments from now. I don't know if we want to create a BaseModel  for them to inherit from. Like:
```python
class ???SubmissionModel(BaseModel):
    """BaseModel for high level DIRAC Workflow"""

    # Allow arbitrary types to be passed to the model
    model_config = ConfigDict(arbitrary_types_allowed=True)

    task: CommandLineTool | Workflow | ExpressionTool
    input_data: Optional[list[str | File] | None] = None

    @field_serializer("task")
    def serialize_task(self, value):
        """Serialize CWL task object to dictionary.

        :param value: CWL task object to serialize.
        :return: Serialized task dictionary.
        :raises TypeError: If value is not a valid CWL task type.
        """
        if isinstance(value, (CommandLineTool, Workflow, ExpressionTool)):
            return save(value)
        else:
            raise TypeError(f"Cannot serialize type {type(value)}")

    @field_serializer("input_data")
    def serialize_input_data(self, value):
        """Serialize an input data list to a list of strings.

        :param value: Input data list to serialize.
        :return: Serialized input data list.
        """
        if value:
            return [save(item) if isinstance(item, File) else item for item in value]
        return None
```

@aldbr do you think it is worth it? Or this will evolve soon and this will not be the case afterward.

src/dirac_cwl/transformation/__init__.py
```diff
@@ -52,6 +54,12 @@ def submit_transformation_client(
     console.print("[blue]:information_source:[/blue] [bold]CLI:[/bold] Validating the transformation...")
     try:
         task = load_document(pack(task_path))
+
+        # Load Transformation inputs if existing
+        input_data = None
+        if inputs_file:
+            input_data = load_inputfile(task.cwlVersion, inputs_file).get("input-data")
```
aldbr: Same here, I am not entirely sure to understand why we need to extract a specific `input-data` key
```suggestion
            input_data = load_inputfile(task.cwlVersion, inputs_file)
```

src/dirac_cwl/transformation/__init__.py
```diff
@@ -103,37 +111,48 @@ def submit_transformation_router(transformation: TransformationSubmissionModel)
     except Exception as exc:
         raise ValueError(f"Invalid DIRAC hints:\n{exc}") from exc
 
+    # Inputs from Transformation inputs_file
+    if transformation.input_data:
```
aldbr: I guess input data should come either from `input_data` or from the query parameters but not both (I would make `if/else` instead of `if/if`).
What do you think?

aldbr (commented):
Since this PR was opened, there has been some changes in the design of the hints so we should be pragmatic and focus on the user interfaces mostly.

The `input_data` dictionary maps **CWL input parameter names** to lists of files. This avoids hardcoding any key name (like `input-data`) and makes the mapping between hint data and workflow inputs explicit.

### CLI convenience: `--inputs-file`

Users can provide a standard CWL inputs YAML file via the CLI:

```bash
dirac-cwl transformation submit workflow.cwl --inputs-file data.yaml
```

Where `data.yaml` is:

```yaml
simulation-files:
  - /lfn/path/file1.root
  - /lfn/path/file2.root
  - /lfn/path/file3.root
```

The CLI reads this file and populates `dirac:ExecutionHooks.input_data` in the hint before submission. This means:

- The router always reads input data from one place (the hint)
- The `--inputs-file` flag is syntactic sugar for populating the hint:

```yaml
hints:
  - class: dirac:ExecutionHooks
    group_size: 5
    input_data:
      simulation-files:
        - /lfn/path/file1.root
        - /lfn/path/file2.root
        - /lfn/path/file3.root
```

- If the hint already contains `input_data` AND `--inputs-file` is provided, emit a warning (`--input_file` should override `input_data`).
- The file format is a standard CWL inputs YAML, nothing new to learn. For simplicity, it should contain only 1 parameter I think (unless CTAO has a use case where they would need multiple static lists of inputs?).

### Dynamic queries (what we already have with `configuration`)

For transformations that discover inputs at runtime (e.g., from upstream transformation outputs), the hint uses a plugin-based configuration instead:

```yaml
hints:
  - class: dirac:ExecutionHooks
    group_size: 5
    input_query:
      plugin: QueryBasedPlugin
      config:
        query_root: "."
        campaign: "pi"
        data_type: "100"
```

A transformation receives dynamic inputs from one upstream transformation only.

### Mutual exclusivity

`input_data` (static) and `input_query` (dynamic) cannot coexist on the same transformation. The system should raise a clear error:

```
Cannot specify both static input data and dynamic input query.
Use input_data (or --inputs-file) for standalone transformations with known files.
Use input_query for transformations that discover inputs from upstream outputs.
```

### Which input gets split?

A transformation receives input data through one file-array input only, whether static (`input_data` with a single key) or dynamic (`input_query` from one upstream transformation). `group_size` applies to that input.

If `input_data` contains multiple keys, raise an error. Multiple dynamic input sources per transformation are not supported, each transformation queries one upstream transformation only.

Non-file inputs (scalars, non-array types) defined in the workflow's regular CWL inputs are passed unchanged to every job.

## Productions

### Static inputs and the first transformation convention

When a production is submitted with `--inputs-file`, the static input data is passed to the first transformation by convention. This follows the natural pipeline structure:

```
t1 (receives static inputs) -> t2 (queries t1 outputs) -> t3 (queries t2 outputs)
```

Downstream transformations typically get their inputs via dynamic queries against the DataCatalog/Bookkeeping service.

In that way we start simple. If it's enough to cover CTAO and LHCb use cases, then it's fine. If we need something more complex (multiple input queries/data), then we can reevaluate based on the use cases.

aldbr:
@arrabito Yes I confirm.
That PR really helped us uncovering many tricky points, thanks @Stellatsuu :slightly_smiling_face: 

As the design of the hint has been moving a lot, I prefer not to spend too much time on that PR and therefore I am going to finish it by adding a `chunk` parameter that will also be integrated to the job submission API that ryuwd is testing in `diracx`.

src/dirac_cwl/production/__init__.py
```diff
@@ -76,6 +76,7 @@
 def submit_production_client(
     task_path: str = typer.Argument(..., help="Path to the CWL file"),
     inputs_file: str | None = typer.Option(None, help="Path to the CWL inputs file"),
+    chunk: str | None = typer.Option(None, help="Split an array input into jobs: PARAM=SIZE (e.g., input-data=3)"),
```
ryuwd, line 79: Me and @chrisburr have decided this is more trouble than it's worth due to various problems with assuming the site location of input files and creating file groups before optimising for this (essentially you risk creating jobs with impossible input data requirements) . It's better left as a detail for the transformation system to handle

ryuwd: I will share a finalised diracx CLI design doc when it stabilises

aldbr (approved):

PR #96, opened 2026-02-04
refactor: dirac-cwl-proto --> dirac-cwl & feat: release-please

ryuwd:


natthan-pigoux (approved):
The overall looks good to me thank you.

.github/workflows/deployment.yml
```diff
@@ -0,0 +1,70 @@
...
+          python -m pip install \
+              build \
+              readme_renderer[md] \
+              setuptools_scm
+      - name: Validate README for PyPI
+        run: |
+          python -m readme_renderer README.md -o /tmp/README.html
+      - name: Build distributions
+        run: |
+          python -m build --outdir $PWD/dist .
```
aldbr: Very last comment (🤞 ), I think it's very good now: you are duplicating that step in `deploy-pypi`, you could potentially upload it as an artifact and download it in `deploy-pypi`. What do you think?

ryuwd: The duplication is a little annoying but necessary in the current setup to ensure we can test the deployment before deploying and get the correct version number from the tag created by release-please by running after the release-please job

PR #97, opened 2026-02-04
feat: LHCb modules migration

### LINKED ISSUE #67: Integrate real LHCb workflow commands (Pre/PostExecution)
We have various LHCbDIRAC workflow modules that would need to be designed differently. Here is the main idea:
- LHCbDIRAC-specific code should go in Pre/PostExecution commands
- LHCb-specific code should mostly go in `lb-prod-run` (in this prototype we will create a wrapper around the existing `lb-prod-run` command as a PoC, and then the goal will be to integrate it to the command itself.

We are going to proceed one command at time probably. The process will consist in:
- disentangling parts of the code that should go in Pre/PostExecution commands, and the `lb-prod-run` wrapper
- once we all agree with the plan, proceed with the implementation
- integrate tests and documentation

### LINKED ISSUE #87: LHCb Workflow: UploadLogFile command
Please read https://github.com/aldbr/dirac-cwl-proto/issues/67 first.

- [x] disentangle the module (https://gitlab.cern.ch/lhcb-dirac/LHCbDIRAC/-/blob/devel/src/LHCbDIRAC/Workflow/Modules/UploadLogFile.py?ref_type=heads)
---
Once you have the green light:
- [ ] implement it
- [ ] unit test it (reuse the existing ones as much as possible, adapting the code to the conventions of the repo https://gitlab.cern.ch/lhcb-dirac/LHCbDIRAC/-/blob/devel/src/LHCbDIRAC/Workflow/Modules/test/Test_Modules_AnalyseXMLSummary.py?ref_type=heads)
- [ ] document it correctly (you can ask for clarification here of course, code is not obvious)

AcquaDiGiorgio:
See #87

First approach, needs review.

I tried to use the current Mock classes (MockDataManager), but I don't think this is the correct way. However, it should not vary much.

I open it as a draft because I still need to create 2 more tests and change the interactions with DIRAC.


AcquaDiGiorgio:
This PR will now tackle the migration of every single module (starting from the ones used in MCSim) by importing LHCb specific functions instead of rewritting them from the ground up

src/dirac_cwl/commands/__init__.py
```diff
@@ -1,5 +1,20 @@
 """Command classes for workflow pre/post-processing operations."""
 
+from .analyze_xml_summary import AnalyseXmlSummary
+from .bookkeeping_report import BookeepingReport
```
aldbr: Oops, missing a `k` 🙂 
```suggestion
from .bookkeeping_report import BookkeepingReport
```

src/dirac_cwl/commands/workflow_accounting.py
```diff
@@ -0,0 +1,108 @@
...
+
+            res = jobStep.checkValues()
+            if not res["OK"]:
+                raise WorkflowProcessingException(
+                    "Values for StepAccounting are wrong:", f"{res['Message']}. Here are the given data: {dataDict}"
+                )
+
+            dsc = DataStoreClient()
+            dsc.addRegister(jobStep)
+            workflow_commons["accounting_registers"] = dsc.__registersList
```
aldbr: I doubt this is going to work as `registersList` is a private attribute.
I guess you would need to add a public `getRegisters()` method within LHCbDIRAC no?

src/dirac_cwl/commands/bookkeeping_report.py
```diff
@@ -0,0 +1,168 @@
...
+            bfilename = f"bookkeeping_{workflow_commons['command_id']}.xml"
+            with open(bfilename, "wb") as bfile:
+                bfile.write(doc)
+
+        except:
+            failed = True
+            raise
+
+        finally:
+            save_workflow_commons(workflow_commons, workflow_commons_path, failed=failed)
```
aldbr: Any reason for not declaring `workflow_commons = {}` before the `try` and keeping `save_workflow_commons` under the `if workflow_commons` here? You might get exception because `workflow_commons` won't be defined I think here. This looks correctly handled in `upload_log_file` for instance

Same issue in `analyze_xml_summary` and `failover_request` and `upload_output_data`

src/dirac_cwl/commands/utils.py
```diff
@@ -0,0 +1,127 @@
...
+import os
+import shutil
+
+from DIRAC import siteName
+from DIRAC.Core.Utilities.ReturnValues import S_ERROR, S_OK
+
+from dirac_cwl.core.exceptions import WorkflowProcessingException
+
+
+def prepare_lhcb_workflow_commons(workflow_commons_path, extra_mandatory_values=[], extra_default_values={}):
```
aldbr: I think it's safer to have immutable default arguments:
```suggestion
def prepare_lhcb_workflow_commons(workflow_commons_path, extra_mandatory_values=None, extra_default_values=None):
```

Here you might have some surprises if you end up calling `prepare_lhcb_workflow_commons` multiple times in a row for any reason. (Just in case if you are not aware: https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil)

src/dirac_cwl/commands/utils.py
```diff
@@ -0,0 +1,127 @@
...
+            json.dump(wf_commons, f)
+    except Exception:
+        os.unlink(wf_file_path)
+        shutil.copy2(wf_backup, wf_file_path)
+        return False
+
+    finally:
+        os.unlink(wf_backup)
+
+    return True
```
aldbr: Looks like the returned `bool` is never checked. Is that expected?

src/dirac_cwl/commands/upload_output_data.py
```diff
@@ -0,0 +1,190 @@
...
+                    "JOB_ID": workflow_commons["job_id"],
+                    "configVersion": workflow_commons["config_version"],
+                    "outputList": workflow_commons["outputs"],
+                    "configName": workflow_commons["config_name"],
+                    "outputDataFileMask": workflow_commons["output_data_file_mask"],
+                }
+                result = constructProductionLFNs(parameters, bk_client)
+
+                if not result["OK"]:
+                    raise WorkflowProcessingException("Unable to construsct production LFNs")
```
aldbr:
```suggestion
                    raise WorkflowProcessingException("Unable to construct production LFNs")
```

src/dirac_cwl/commands/failover_request.py
```diff
@@ -0,0 +1,114 @@
...
+                if file_report.getFiles():
+                    result = file_report.generateForwardDISET()
+                    if result["OK"] and result["Value"]:
+                        request.addOperation(result["Value"])
+
+                job_report.setApplicationStatus("Job Finished Successfully", True)
+
+            self.generateFailoverFile(job_report, request, workflow_commons)
+
+        except:
```
aldbr: Isn't it too wide? (you can catch `KeyboardInterrupt` for instance here)
```suggestion
        except Exception:
```

src/dirac_cwl/commands/upload_log_file.py
```diff
@@ -0,0 +1,149 @@
...
+
+            ops = Operations()
+            log_se = ops.getValue("LogStorage/LogSE", "LogSE")
+            log_extensions = ops.getValue("LogFiles/Extensions", [])
+
+            _prepareRequest(request, workflow_commons["job_id"])
+            failover_transfer = FailoverTransfer(request)
+            job_report = JobReport(workflow_commons["job_id"])
+
+            res = systemCall(0, shlex.split("ls -al"))
```
aldbr: Don't you need to display the output?

src/dirac_cwl/commands/upload_log_file.py
```diff
@@ -0,0 +1,149 @@
...
+            result = _zip_files(workflow_commons["prod_job_id"], selectedFiles)
+            if not result["OK"]:
+                job_report.setApplicationStatus("Failed to create zip of log files")
+                return
+
+            zip_file_name = result["Value"]
+
+            # Instantiate the failover transfer client with the global request object
+            if not failover_transfer:
+                failover_transfer = FailoverTransfer(request)
```
aldbr: Looks like `failover_transfer` is already defined above, is that expected?
```suggestion
```

src/dirac_cwl/commands/workflow_accounting.py
```diff
@@ -0,0 +1,108 @@
...
+            else:
+                xf_o = _generate_xml_object(
+                    workflow_commons["cleaned_application_name"],
+                    workflow_commons["production_id"],
+                    workflow_commons["prod_job_id"],
+                    workflow_commons["command_number"],
+                    workflow_commons["command_id"],
+                )
+
+            now = datetime.utcnow()
```
aldbr: `datetime.utcnow()` is deprecated, `datetime.now(timezone.utc)` should be favored.

src/dirac_cwl/commands/upload_output_data.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        fail = False
+        try:
+            workflow_commons_path = kwargs.get("workflow_commons_path", os.path.join(job_path, "workflow_commons.json"))
+
+            workflow_commons = prepare_lhcb_workflow_commons(
+                workflow_commons_path,
+                extra_mandatory_values=["output_data_step", "output_SEs"],
+                extra_default_values={
+                    "file_descendants": None,
```
aldbr: ?
```suggestion
                    "file_descendents": None,
```

src/dirac_cwl/commands/upload_output_data.py
```diff
@@ -0,0 +1,190 @@
...
+            final = _resolveSEs(
+                file_metadata,
+                None,
+                workflow_commons["site_name"],
+                workflow_commons["output_mode"],
+                workflow_commons["run_number"],
+            )
+
+            if workflow_commons["inputs"]:
+                lfns_with_descendants = workflow_commons["file_descendants"]
```
aldbr: Same here no?
```suggestion
                lfns_with_descendents = workflow_commons["file_descendants"]
```

src/dirac_cwl/commands/analyze_xml_summary.py
```diff
@@ -0,0 +1,84 @@
...
+
+            file_report = FileReport()
+            job_report = JobReport(workflow_commons["job_id"])
+
+            file_report.statusDict = workflow_commons["file_report_files_dict"]
+
+            jobOk = _isXMLSummaryOK(xf_o)
+
+            if jobOk:
+                jobOk = _areInputsOK(
```
aldbr: Note: in CWL, I think the new Rust implementation of `lb-prod-run` is going to be used: 

And there are some checks around the xml summary directly integrated into it: https://gitlab.cern.ch/roneil/lbprodrun-rs/-/blob/main/src/summary.rs

So I assume we could only report the file status of the problematic files in this command

src/dirac_cwl/commands/utils.py
```diff
@@ -0,0 +1,127 @@
...
+        "prod_job_id",
+        "number_of_events",
+        "application_name",
+        "application_version",
+        "inputs",
+        "outputs",  # outputList
+        "executable",
+        "command_id",  # StepID
+        "command_number",
+    ]
```
aldbr: I think here it would be better to have a pydantic model to get types and potentially some validation

src/dirac_cwl/commands/utils.py
```diff
@@ -0,0 +1,127 @@
...
+    for value in mandatory_values:
+        if value not in workflow_commons:
+            missing_values.append(value)
+
+    if missing_values:
+        raise WorkflowProcessingException(
+            f"The following values are missing in workflow_commons.json: {missing_values}"
+        )
+
+    commons_defaults = {
```
aldbr: Are you going to extract that from the CWL file?
We also want that as part of a pydantic model

src/dirac_cwl/commands/utils.py
```diff
@@ -0,0 +1,127 @@
...
+def save_workflow_commons(wf_commons, wf_file_path, request=None, failed=False):
+    """Update the workflow_commons file to accomodate for the new values.
+
+    Ensures that no data is lost during the update by creating a backup.
+    """
+    if not (os.path.exists(wf_file_path) and os.path.isfile(wf_file_path)):
+        raise WorkflowProcessingException(f"Workflow Commons file '{wf_file_path}' not found")
+
+    wf_filename = os.path.basename(wf_file_path)
+    wf_backup = f"{wf_filename}.bak"
```
aldbr: Don't we want to store `wf_backup` within `job_path`?

src/dirac_cwl/commands/analyze_xml_summary.py
```diff
@@ -0,0 +1,84 @@
...
+                workflow_commons_path,
+                extra_mandatory_values=[
+                    "bk_step_id",
+                ],
+                extra_default_values={
+                    "bookkeeping_LFNs": [],
+                    "size": {},
+                    "md5": {},
+                    "guid": {},
+                    "sim_description": "NoSimConditions",
```
aldbr: These extra values should also be structured within a pydantic model ideally

src/dirac_cwl/commands/bookkeeping_report.py
```diff
@@ -0,0 +1,168 @@
...
+                "cputime": cputime,
+                "numberOfProcessors": number_of_processors,
+                "production_id": workflow_commons["production_id"],
+                "jobID": workflow_commons["job_id"],
+                "siteName": workflow_commons["site_name"],
+                "jobType": workflow_commons["job_type"],
+                "applicationName": workflow_commons["application_name"],
+                "applicationVersion": workflow_commons["application_version"],
+                "numberOfEvents": workflow_commons["number_of_events"],
+            }
```
aldbr: Same here

src/dirac_cwl/commands/workflow_accounting.py
```diff
@@ -0,0 +1,108 @@
...
+            if "xml_summary_path" in workflow_commons:
+                xf_o = XMLSummary(workflow_commons["xml_summary_path"])
+            else:
+                xf_o = _generate_xml_object(
+                    workflow_commons["cleaned_application_name"],
+                    workflow_commons["production_id"],
+                    workflow_commons["prod_job_id"],
+                    workflow_commons["command_number"],
+                    workflow_commons["command_id"],
+                )
```
aldbr: Is this expected? In the original file, it looks like there is an exception raised if `xf_o` is not defined. Am I wrong?

src/dirac_cwl/commands/workflow_accounting.py
```diff
@@ -0,0 +1,108 @@
...
+                "InputData": sum(xf_o.inputFileStats.values()),
+                "OutputData": sum(xf_o.outputFileStats.values()),
+                "InputEvents": xf_o.inputEventsTotal,
+                "OutputEvents": xf_o.outputEventsTotal,
+            }
+
+            jobStep.setValuesFromDict(dataDict)
+
+            res = jobStep.checkValues()
+            if not res["OK"]:
```
aldbr: To avoid writing `S_OK()/S_ERROR()` you can use `returnValueOrRaise()` from DIRAC 🙂

src/dirac_cwl/commands/workflow_accounting.py
```diff
@@ -0,0 +1,108 @@
...
+                    "Values for StepAccounting are wrong:", f"{res['Message']}. Here are the given data: {dataDict}"
+                )
+
+            dsc = DataStoreClient()
+            dsc.addRegister(jobStep)
+            workflow_commons["accounting_registers"] = dsc.__registersList
+
+        except Exception as e:
+            failed = True
+            raise WorkflowProcessingException() from e
```
aldbr: No message here?

src/dirac_cwl/commands/workflow_accounting.py
```diff
@@ -0,0 +1,108 @@
...
+                    workflow_commons["prod_job_id"],
+                    workflow_commons["command_number"],
+                    workflow_commons["command_id"],
+                )
+
+            now = datetime.utcnow()
+            jobStep.setStartTime(now)
+            jobStep.setEndTime(now)
+
+            dataDict = {
```
aldbr: This also needs to be typed

src/dirac_cwl/commands/utils.py
```diff
@@ -0,0 +1,127 @@
...
+        "options_file": None,
+        "options_line": None,
+        "extra_packages": "",
+        "multi_core": False,
+        "max_number_of_processors": None,
+        "system_config": None,
+        "mcTCK": None,
+        "condDB_tag": None,
+        "DQ_tag": None,
+        "step_status": S_OK(),
```
aldbr: `step_status` is either `Done` or `Failed`, not `S_OK/S_ERROR`. Check within the `StepAccounting` module in LHCbDIRAC
```suggestion
        "step_status": "Done",
```

aldbr: At some point that should just be a boolean but we cannot change it for now (though you can add a comment for later)

src/dirac_cwl/commands/utils.py
```diff
@@ -0,0 +1,127 @@
...
+        "job_type",
+        "production_id",
+        "prod_job_id",
+        "number_of_events",
+        "application_name",
+        "application_version",
+        "inputs",
+        "outputs",  # outputList
+        "executable",
+        "command_id",  # StepID
```
aldbr: I am not sure to understand why this is renamed `command_id` 😅

src/dirac_cwl/commands/failover_request.py
```diff
@@ -0,0 +1,114 @@
...
+from DIRAC.WorkloadManagementSystem.Client.JobReport import JobReport
+from LHCbDIRAC.Workflow.Modules.FailoverRequest import _prepareRequest
+
+from dirac_cwl.core.exceptions import WorkflowProcessingException
+
+from .core import PostProcessCommand
+from .utils import prepare_lhcb_workflow_commons, save_workflow_commons
+
+
+class FailoverRequest(PostProcessCommand):
```
aldbr: It looks like you are not setting the application status, is that expected? (this can be done through diracX)

aldbr (changes requested):

src/dirac_cwl/commands/workflow_accounting.py
AcquaDiGiorgio: I can't believe I didn't know that python actually cared about protected/private attributes 😐 
I think I will add the `jobStep` to the `workflow_commons` directly, instead of adding it to the `DataStore` registers list and then extracting that list.

src/dirac_cwl/commands/__init__.py
AcquaDiGiorgio: 🤦‍♂️

src/dirac_cwl/commands/workflow_accounting.py
AcquaDiGiorgio: It finishes gracefully without raising. However, as we cannot store the XmlSummary object in `workflow_commons`, I guess we could just stop the execution instead of trying to generate it.

src/dirac_cwl/commands/utils.py
AcquaDiGiorgio: In the [tests](https://gitlab.cern.ch/lhcb-dirac/LHCbDIRAC/-/blob/devel/src/LHCbDIRAC/Workflow/Modules/test/Test_Modules_StepAccounting.py?ref_type=heads#L87-97), the exectue function of every Module sets up `stepStatus=S_OK()` or `stepStatus=S_ERROR()`. The same goes for the attibute `stepStatus` in [`ModuleBase`](https://gitlab.cern.ch/lhcb-dirac/LHCbDIRAC/-/blob/devel/src/LHCbDIRAC/Workflow/Modules/ModuleBase.py?ref_type=heads#L994).

Should I change it to `Done`/`Failed` then?

src/dirac_cwl/commands/failover_request.py
AcquaDiGiorgio: What do you mean?
What application status is not being set?

src/dirac_cwl/commands/failover_request.py
AcquaDiGiorgio: It is, it also should raise `WorkflowException` from the `Exception` orginally raised

src/dirac_cwl/commands/upload_log_file.py
AcquaDiGiorgio: I still need to add the logging, but before I wanted to know if this was how you wanted to proceed with this PR.

src/dirac_cwl/commands/utils.py
AcquaDiGiorgio: Future logging purposes, at the moment it does nothing. 
Should the workflow fail if it is unable to save the commons?

src/dirac_cwl/commands/bookkeeping_report.py
AcquaDiGiorgio: I missed it :/

src/dirac_cwl/commands/upload_output_data.py
```diff
@@ -0,0 +1,196 @@
...
+            logger.info("Sending BK record:\n%s", bkXML)
+            try:
+                returnValueOrRaise(_sendBKReport(workflow_commons.bk_client, workflow_commons.request, bkXML))
+                logger.info("Bookkeeping report sent for %s", bk_file)
+            except SErrorException as e:
+                logger.error("Could not send Bookkeeping XML file to server:\n%s", e)
+                logger.info("Preparing DISET request for %s", bk_file)
+
+        logger.info("Creating DISABLE_WATCHDOG_CPU_WALLCLOCK_CHECK in order to disable the Watchdog")
+        with open("DISABLE_WATCHDOG_CPU_WALLCLOCK_CHECK", "w") as f:
```
aldbr: When you create new files like this, please make sure they are created in `job_path`.
In the PushJobAgent, we might have multiple job wrappers running in parallel in the future, you want to make sure they will not collide.

```suggestion
        with open(job_path / "DISABLE_WATCHDOG_CPU_WALLCLOCK_CHECK", "w") as f:
```

src/dirac_cwl/commands/upload_output_data.py
```diff
@@ -0,0 +1,196 @@
...
+from .core import PostProcessCommand
+from .workflow_commons import StepStatus, WorkflowCommons
+
+logger = logging.getLogger(__name__)
+
+
+class UploadOutputData(PostProcessCommand):
+    """Registers every output generated to the corresponding SE and Catalog or to the FailoverSE in case of failure."""
+
+    def _execute(self, job_path: os.PathLike, workflow_commons: WorkflowCommons, **kwargs) -> None:
```
aldbr: Out of curiosity, why `os.PathLike` here and not `pathlib.Path`?

src/dirac_cwl/commands/report_bookkeeping.py
```diff
@@ -0,0 +1,162 @@
...
+        # Generate SimulationConditions
+        if step_commons.application_name == "Gauss":
+            job_info.simulation_condition = workflow_commons.sim_description
+
+        # Convert job_info object to XML
+        doc = job_info.to_xml()
+
+        # Write to file
+        bfilename = f"bookkeeping_{step_commons.id}.xml"
+        with open(bfilename, "wb") as bfile:
```
aldbr, line 165: For instance, here you also want this to be part of `job_path`, let's write it explicitly may be.
```suggestion
        with open(job_path / bfilename, "wb") as bfile:
```

src/dirac_cwl/commands/report_bookkeeping.py
```diff
@@ -0,0 +1,162 @@
...
+import copy
+import logging
+import os
+from typing import Any, Dict
+
+from DIRAC.Core.Utilities.ReturnValues import SErrorException, returnValueOrRaise
+from DIRAC.Workflow.Utilities.Utils import getStepCPUTimes
+from LHCbDIRAC.Core.Utilities.ProductionData import constructProductionLFNs
+from LHCbDIRAC.Workflow.Modules.BookkeepingReport import (
+    _generate_xml_object,
```
aldbr, line 14: You are still extracting protected functions here 😅

src/dirac_cwl/commands/upload_log_file.py
```diff
@@ -0,0 +1,173 @@
...
+        log_extensions = ops.getValue("LogFiles/Extensions", [])
+
+        _prepareRequest(workflow_commons.request, workflow_commons.job_id)
+
+        try:
+            file_list = returnValueOrRaise(systemCall(0, shlex.split("ls -al")))
+        except SErrorException as e:
+            logger.error("Failed to list the log directory\n%s", e)
+
+        if file_list:
```
aldbr: What happens if an exception is raised? I assume `file_list` would not exist and you would get another exception?
Can you test that please?

src/dirac_cwl/commands/workflow_commons.py
```diff
@@ -0,0 +1,301 @@
...
+        return self._dsc
+
+    def generate_failover_file(self):
+        """Create a request.json file."""
+        try:
+            diset_op = returnValueOrRaise(self.job_report.generateForwardDISET())
+        except SErrorException as e:
+            self._logger.warning("Could not generate Operation for job report", exc_info=e)
+
+        if diset_op:
```
aldbr: Same issue here. Please test it if you can

src/dirac_cwl/commands/workflow_commons.py
```diff
@@ -0,0 +1,301 @@
...
+        self.request_dict = json.loads(self.request.toJSON()["Value"])
+        self.accounting_registers = self.dsc._DataStoreClient__registersList
+
+        try:
+            wf_dict = self.model_dump(mode="json")
+            with open(wf_path, "w", encoding="utf-8") as f:
+                json.dump(wf_dict, f)
+        except Exception as e:
+            self._logger.exception("Failed to save the workflows commons in a file", exc_info=e)
+            raise
```
aldbr, line 191: If `model_dump` raises an exception, then I assume you want to restore the backup, but here you delete it in the `finally`. So effectively the backup does not do its job IIUC 😅

src/dirac_cwl/commands/workflow_commons.py
```diff
@@ -0,0 +1,301 @@
...
+
+    bookkeeping_lfns: list[str] = []
+    prod_output_lfns: list[str] = []
+
+    file_descendents: list[str] = []
+    file_report_files_dict: dict = {}
+    accounting_registers: list = []
+    xml_summary_paths: dict[str, str] = {}
+    request_dict: dict = {}
+    files_report_files_dict: dict = {}
```
aldbr: Looks like `file(s)_report_files_dict` appear twice

```suggestion
    file_report_files_dict: dict = {}
    accounting_registers: list = []
    xml_summary_paths: dict[str, str] = {}
    request_dict: dict = {}
```

src/dirac_cwl/commands/workflow_accounting.py
```diff
@@ -0,0 +1,92 @@
...
+
+from dirac_cwl.core.exceptions import WorkflowProcessingException
+
+from .core import PostProcessCommand
+from .workflow_commons import Step, WorkflowCommons
+
+logger = logging.getLogger(__name__)
+
+
+class WorkflowAccounting(PostProcessCommand):
```
aldbr: Minor: can we try to always name the class the same way? Likely something like <verb>_<name>.
For instance having module names like: `analyze_summary`, `report_bookkeeping`, `upload_output_data`, `register_accounting_report`. 
And therefore class names following the same logic: `RegisterAccountingReport`.

Any idea? What do you think? I don't have any strong opinion about the names, I just want to have something consistent.

src/dirac_cwl/commands/workflow_accounting.py
```diff
@@ -0,0 +1,92 @@
...
+        if step_commons.start_stats:
+            cpu_times["StartStats"] = step_commons.start_stats
+
+        if not step_commons.application_name:
+            logger.info("Not an application step: it will not be accounted")
+            return
+
+        exec_time, cpu_time = getStepCPUTimes(cpu_times)
+
+        cpu_norm_factor = gConfig.getValue("/LocalSite/CPUNormalizationFactor", 0.0)
```
aldbr: Since we are going to transition to DiracX, we want to rename as much as possible `cpu_norm_factor` as `cpu_power` (it's been years we are slowly transitioning to that "semantics").

```suggestion
        cpu_power = gConfig.getValue("/LocalSite/CPUNormalizationFactor", 0.0)
```

src/dirac_cwl/commands/workflow_accounting.py
```diff
@@ -0,0 +1,92 @@
...
+        norm_cpu = cpu_time * cpu_norm_factor
```
aldbr: In the current workflow module, this is gated with `if cpu_norm_factor`, and `cpu_time` is used in the case `cpu_power` (`cpu_norm_factor`) is not available (which should be really really rare).

aldbr: I am wondering whether there is a way in CWL to get some info about the workflow steps. If that is the case, then may be you could try to extract that as a `JobStep`. I know that Ryan managed to extract some information about the workflow steps in his PR, but I haven't checked how. This can probably come as a second step, but worth keeping in mind!

src/dirac_cwl/commands/failover_request.py
```diff
@@ -0,0 +1,71 @@
...
+        except SErrorException as e:
+            logger.error("Something went wrong trying fileReport.commit() %s", e)
+
+        if workflow_commons.file_report.getFiles():
+            logger.error("On first attempt, failed to report file status to TransformationDB")
+            try:
+                value = returnValueOrRaise(workflow_commons.file_report.generateForwardDISET())
+                if not value:
+                    logger.info("On second attempt, files correctly reported to TransformationDB")
+                elif workflow_commons.step_status == StepStatus.Done:
```
aldbr: Also based on my previous comment in the `WorkflowAccounting`: it would be interesting to see whether there is a way to get the status of the cwl execution vs a step execution.

Because if this is possible, then may be there is a way to reproduce exactly what we have in the workflow modules with the conditions like `if workflowStatus and jobStatus...`

src/dirac_cwl/commands/workflow_accounting.py
```diff
@@ -0,0 +1,92 @@
...
+
+        try:
+            returnValueOrRaise(job_step.checkValues())
+        except SErrorException as e:
+            logger.error("Values for StepAccounting are wrong: Here are the given data: %s", data_dict, exc_info=e)
+            raise WorkflowProcessingException(
+                f"Values for StepAccounting are wrong. Here are the given data: {data_dict}"
+            ) from e
+
+        workflow_commons.dsc.addRegister(job_step)
```
aldbr: Are you sure it works?

src/dirac_cwl/commands/workflow_commons.py
```diff
@@ -0,0 +1,301 @@
...
+    step_status: StepStatus = StepStatus.Done
+
+    # Private attributes
+    _request = PrivateAttr(default=None)
+    _failover_request = PrivateAttr(default=None)
+    _job_report = PrivateAttr(default=None)
+    _file_report = PrivateAttr(default_factory=FileReport)
+    _data_manager = PrivateAttr(default_factory=DataManager)
+    _bk_client = PrivateAttr(default_factory=BookkeepingClient)
+    _dsc = PrivateAttr(default_factory=DataStoreClient)
```
aldbr: I would not carry the clients, I would only add the clients in the commands where needed. Why did you decide to have all them here?
```suggestion
```

The `jobReport` could potentially be part of the signature of pre/post process commands though, because it's also used in the `JobWrapper` (we could imagine having it reporting the "application status" being the name of the pre/post process command - that would be in the base classes - and passing it to the commands so that they can influence the status of the job if needed).

src/dirac_cwl/commands/workflow_commons.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+class StepStatus(str, Enum):
+    """Workflow status."""
+
+    Done = "Done"
+    Failed = "Failed"
+
+
+class Step(BaseModel):
```
aldbr, line 29: There might be a way of having one step per CWL workflow step.
Check the workflow examples we have here (this is not MCSim but it's very similar): https://gitlab.cern.ch/lhcb-dpa/analysis-productions/lbapcommon/-/blob/master/tests/example_workflows/complex_workflow_with_filtering/AnaProd_example_workflows_job_tuple%2Csplit%2Cfilter_B0.cwl?ref_type=heads

test/test_commands.py
aldbr, line 1: Could you split that file into multiple smaller files please? One per command

aldbr (changes requested):

src/dirac_cwl/commands/upload_output_data.py
AcquaDiGiorgio: As far as I'm aware, the recommended way of typing path parameters is by using `os.PathLike`, which reminds me that that should be `Union[str, os.PathLike]`, as `str` is not a subtype of `os.PathLike`.
On the other hand, I mostly use `pathlib.Path` for file operations, so maybe is not a bad idea to just use pathlib.

src/dirac_cwl/commands/report_bookkeeping.py
AcquaDiGiorgio: Those protected functions are the ones I developed in my LHCbDIRAC fork :) 
Do you want me to set them as public?

Also, to access the `DataStoreClient` accounting registers private variable I used `dsc._DataStoreClient__registersList`. This is works because I use the mangled named. Is a dirty solution, but I don't want to touch LHCbDIRAC's `DataStoreClient` for this.

src/dirac_cwl/commands/upload_log_file.py
AcquaDiGiorgio: That's true, I missed that. I will check every newly added `returnValueOrRaise`, as it changes the code blocks.

src/dirac_cwl/commands/workflow_accounting.py
AcquaDiGiorgio: Okay, so you mean having something like `{verb}_{noun}[_{noun}]*`
I mean, I prefer a consistent convention and it also more human readable. 

This is very small, so I'm okay with this change.

src/dirac_cwl/commands/workflow_accounting.py
AcquaDiGiorgio: > In the current workflow module, this is gated with if cpu_norm_factor, and cpu_time is used in the case cpu_power (cpu_norm_factor) is not available (which should be really really rare).

I wanted to put gConfig.getValue("...", 1.0), but it seems I forgot. This way, we don't need to check anything, just multiply the values and that's it.

>  I know that Ryan managed to extract some information about the workflow steps in his PR, but I haven't checked how. This can probably come as a second step, but worth keeping in mind!

I haven't looked at his DiracX PR in too much detail, but I guess that the execution time could be measured in the `DiracExecutor`, which implements cwl-tools's `SingleJobExecutor`.

src/dirac_cwl/commands/workflow_accounting.py
AcquaDiGiorgio: Why wouldn't it?
When the command execution ends, the registers are extracted from the client and stored for later use.

src/dirac_cwl/commands/workflow_commons.py
AcquaDiGiorgio: My idea was to maintain the same structure and keep the code on each command as clean as possible. 
However, I understand that instantiating all clients when most are not needed is unefficient.

> The jobReport could potentially be part of the signature of pre/post process commands though, because it's also used in the JobWrapper

This could be a nice addition. Also, Ryan's job_wrapper creates a `DataManager`. We could give the possibility of setting the clients in `CommandBase` and if they are not present while executing the command, just instantatiate them.

src/dirac_cwl/commands/workflow_commons.py
AcquaDiGiorgio: Okay, I will look into this

src/dirac_cwl/commands/upload_output_data.py
aldbr: Ok I didn't know. From what I read, `os.PathLike[str]` is abstract and useful in your specific context for `typing`, but then you can (should?) "cast" it as a `pathlib.Path` to manipulate it.
Now given that we completely control this area of the code, I naively assume that having `pathlib.Path` is safe. 
But I would happily follow your recommendation :slightly_smiling_face:

src/dirac_cwl/commands/report_bookkeeping.py
aldbr: It's likely not recommended but probably fine here

PR #127, opened 2026-04-02
feat: replace plugin system with commands and dirac:Job hint

ryuwd:
- Replace ExecutionHooksBasePlugin/QueryBasedPlugin/registry with the commands system. 

- run_job() now reads the dirac:Job hint via JobHint from diracx-core and builds StoreOutputCommand for output storage.

- Commands are now async. 

- convert_to_jdl() deprecated in favour of diracx-logic. 

- Job wrapper template fetches CWL and params from diracX API instead of reading job.json from sandbox.

### DIFF
--- a/docs/design/cwl-input-sandbox-design.md
+++ b/docs/design/cwl-input-sandbox-design.md
@@ -1,0 +1,121 @@
+# Design: Input Sandbox via Replica Map
+
+## Context
+
+The `dirac:Job` hint declares input sandbox files via `input_sandbox` source references:
+
+```yaml
+hints:
+  - class: dirac:Job
+    input_sandbox:
+      - source: helper_script
+      - source: config_files
+        path: "conf/"
+```
+
+These reference CWL inputs whose values are files that need to be available on the worker node. The sandbox store handles upload/download of tar archives.
+
+## Design
+
+### Client Side (before submission)
+
+1. User uploads sandbox files via the sandbox store API → receives a sandbox ID (e.g., `SB:SandboxSE|/S3/diracx-sandbox-store/sha256:abc123.tar.zst`)
+2. User references the sandbox in their input YAML using a `SB:` prefixed path:
+
+```yaml
+# input.yaml
+helper_script:
+  class: File
+  path: "SB:SandboxSE|/S3/diracx-sandbox-store/sha256:abc123.tar.zst:helper.sh"
+
+config_files:
+  - class: File
+    path: "SB:SandboxSE|/S3/diracx-sandbox-store/sha256:abc123.tar.zst:config/app.yaml"
+  - class: File
+    path: "SB:SandboxSE|/S3/diracx-sandbox-store/sha256:abc123.tar.zst:config/db.yaml"
+```
+
+The format is `SB:<sandbox_id>:<relative_path_inside_tar>`.
+
+### Submission (diracx-logic)
+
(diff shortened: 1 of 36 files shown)


docs/design/cwl-input-sandbox-design.md
```diff
@@ -0,0 +1,121 @@
...
+### Client Side (before submission)
+
+1. User uploads sandbox files via the sandbox store API → receives a sandbox ID (e.g., `SB:SandboxSE|/S3/diracx-sandbox-store/sha256:abc123.tar.zst`)
+2. User references the sandbox in their input YAML using a `SB:` prefixed path:
+
+```yaml
+# input.yaml
+helper_script:
+  class: File
+  path: "SB:SandboxSE|/S3/diracx-sandbox-store/sha256:abc123.tar.zst:helper.sh"
```
natthan-pigoux, line 29: I'm wondering if we shouldn't use the [File](https://www.commonwl.org/v1.2/CommandLineTool.html#File) `location:` instead of `path:`  to provide the sandbox url and keep path just with a local path.

Since the CWL definition says `The location may refer to a local or remote resource;`  it seems closer to what is used here. What do you think?

We could think of something similar for the input/output data as well.

ryuwd: Great point! I am fixing it right now in DIRACGrid/diracx#877

src/dirac_cwl/job/job_wrapper.py
```diff
@@ -263,98 +260,61 @@ async def post_process(
...
-                status=JobStatus.COMPLETING, minor_status=JobMinorStatus.OUTPUT_DATA_UPLOADED
+    def _build_commands_from_hint(self, job_hint: JobHint) -> None:
+        """Build pre/post-process commands from the dirac:Job hint.
+
+        The ``type`` field determines which commands are attached.
+        I/O config from the hint is used to configure commands.
+        """
+        # Extract I/O config from the hint
+        output_paths = {
+            entry.source: entry.output_path for entry in job_hint.output_data
```
natthan-pigoux, line 304: Does that handle the case of glob outputs?

We have a use case in CTAO wf like:
```
outputs:
  grid_output_files:
    type: File[]
    doc: Output files including sim_telarray output and log files.
    outputBinding:
      glob: simtools-grid-output/*
```
where we want `grid_output_files` to be an output_data, where the files from the glob need to become LFNs, so be kind of prefixed by what would be in:
```
hints:
 - class: dirac:Job
    output_data:
      - source: grid_output_files
        path: "/some/fc/path/"
```
I'm not sure this is the case here since we append one by one the `output_path` in the postprocess command.

(tell me if it's not clear)

ryuwd: It is designed to work this way yes 🤞 

`cwltool` will collect the outputs of the workflow and emit a JSON blob to stdout which is read by the diracx JobWrapper and used to upload the data. The PostProcessCommand for doing that (the upload) isn't in place yet but the wiring is mostly done

ryuwd: You will need to make sure the output file(s) `id` matches the output you have designated in the hint

natthan-pigoux: Ok great ! Thanks for the precision :+1:

docs/design/cwl-input-sandbox-design.md
ryuwd: Feel free to leave comments on the PR DIRACGrid/diracx#877 as there you will find the most up to date implementation details

PR #130, opened 2026-07-14
refactor: use cwltool loader for standardized cwl doc handling

### LINKED ISSUE #108: [CWL Executor] use cwl_utils to print the workflow
Here was a comment, which I think was not addressed in the previous PR:
```
Out of curiosity, have you noticed that cwltool was providing a command to print a dot representation of the workflow?
I wonder whether printing and displaying this dot file using a python library would simplify the function.
```

_Originally posted by @aldbr in https://github.com/DIRACGrid/dirac-cwl/pull/94#discussion_r2774463938_
            

  ## Context
 
  The executor CLI has a `--print-workflow` flag that prints a human-readable summary
  of the workflow into the job log before execution (steps in order, each step's
  inputs with their sources, its outputs, plus workflow-level inputs/outputs). This is
  mainly for job-log forensics: understanding what ran without opening the CWL file.
 
  ## Problem
  
  `print_workflow_visualization()` in `src/dirac_cwl/job/executor/__main__.py` builds
  this view by `yaml.safe_load`-ing the raw CWL file and hand-parsing it. It has to
  handle both authoring styles (mapping vs list) for inputs/steps/outputs manually,
  and it silently mis-renders anything it doesn't know: packed `$graph` documents,
  subworkflows, scatter. It duplicates CWL document normalization that cwltool
  already does correctly.

  Note: the originally suggested `cwltool --print-dot` is **not** a drop-in fix. Its
  DOT output only contains step-to-step edges, with no port names, types, or labels,
  i.e. less information than the current view. Use cwltool's *loader* instead.

  ## Proposed approach
  
  Keep the rich rendering; replace the parsing. Load the workflow through cwltool and
  walk the normalized document (after loading, inputs/steps/outputs are always lists
  of dicts with absolute IDs, all the dict/list branches can be deleted):
 
  ## Definition of done
  
  - [ ] print_workflow_visualization no longer parses the CWL YAML itself; the
  structure comes from cwltool's loader.
  - [ ] The rendered output still shows: cwlVersion + doc, workflow inputs
  (name, type, label), steps in order with per-step inputs (name ← source)
  and outputs, and workflow outputs (name, type, outputSource).
  Complex types (arrays, records) are rendered best-effort.
  - [ ] Renders correctly for: mapping-style CWL, list-style CWL, and a workflow with
  inline run: tools. A CommandLineTool (no steps) gets a graceful
  inputs/outputs-only view instead of an error.
  - [ ] A unit test exercises the function on at least one workflow fixture and
  asserts on the rendered content.
  - [ ] No new dependencies.

  ## Out of scope

  - ASCII 2-D graph layout or any interactive TUI (no viable pure-Python library?)
  - Generating SVG/PNG. Users who want a picture can run `cwltool --print-dot workflow.cwl | dot -Tsvg > wf.svg` — at most, mention this in the docs.

amaltaro:
Fixes #108

With this pull request, we adopt the cwltool loader functionality to normalize and handle CWL documents, instead of manually juggling with fields and different data types.

Tested with `pixi run dirac-cwl-run test/workflows/crypto/description.cwl --print-workflow`, where exactly the same workflow visualization is retained.
Also added some unit tests that can be triggered with `pixi run pytest test/test_job_executor.py`

### DIFF
--- a/src/dirac_cwl/job/executor/__main__.py
+++ b/src/dirac_cwl/job/executor/__main__.py
@@ -76,7 +76,7 @@ def print_workflow_visualization(workflow_path: Path):
         )
 
         # Show basic info
-        cwl_version = getattr(cwl, "cwlVersion", "Unknown")
+        cwl_version = cwl.cwlVersion
         description = getattr(cwl, "doc", getattr(cwl, "label", ""))
 
         console.print("[bold cyan]CWL Version:[/bold cyan]", cwl_version)
@@ -133,7 +133,7 @@ def print_workflow_visualization(workflow_path: Path):
 
             console.print()
 
-        # Show final outputs (handle both dict and list formats)
+        # Show final outputs
         if outputs:
             console.print("[bold magenta]📤 FINAL OUTPUTS:[/bold magenta]")
             for out_item in outputs:
--- a/src/dirac_cwl/job/executor/__main__.py
+++ b/src/dirac_cwl/job/executor/__main__.py
@@ -26,6 +26,11 @@ def _get_package_version(package: str) -> str:
         return "unknown"
 
 
+def _strip_prefix(x: str) -> str:
+    """Utilitarian function to strip .# prefix from a string."""
+    return x.replace(".#", "")
+
+
 # Create Typer app with context settings to allow extra arguments
 app = typer.Typer(context_settings={"allow_extra_args": True, "ignore_unknown_options": True})
 
@@ -52,12 +57,16 @@ def version_callback(value: bool):
 
 def print_workflow_visualization(workflow_path: Path):
     """Print a nice visualization of the workflow structure with graph representation."""
-    import yaml
+    from cwl_utils.pack import pack
+    from cwl_utils.parser import load_document
 
     try:
-        with open(workflow_path, "r") as f:
-            cwl = yaml.safe_load(f)
+        cwl = load_document(pack(str(workflow_path)), baseuri=".")
+    except Exception as e:
+        console.print(f"[red]⚠ Could not load workflow:[/red] {e}\n")
+        raise typer.Exit(1) from e
 
+    try:
         console.print()
         console.print(
             Panel.fit(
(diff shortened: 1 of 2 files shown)


src/dirac_cwl/job/executor/__main__.py
```diff
@@ -67,146 +76,97 @@ def print_workflow_visualization(workflow_path: Path):
         )
 
         # Show basic info
-        cwl_version = cwl.get("cwlVersion", "Unknown")
-        doc = cwl.get("doc", cwl.get("label", ""))
+        cwl_version = getattr(cwl, "cwlVersion", "Unknown")
```
aldbr: Just wondering if `getattr` is not misleading in this context, as I think the `cwl` instance you get will always have `cwlVersion`, `doc`, `inputs`, .... To double check, but I think they just exist with default value like `None` or `[]` if nothing is provided.

This would likely work:
```suggestion
        cwl_version = cwl.cwlVersion # default value is likely 1.2, to double check
        cwl_version = cwl.cwlVersion or "Unknown" # if there is no default value
```

src/dirac_cwl/job/executor/__main__.py
```diff
@@ -67,146 +76,97 @@ def print_workflow_visualization(workflow_path: Path):
...
+                for step_out in getattr(step_spec, "out", []):
+                    step_out_name = _strip_prefix(step_out).replace(step_name + "/", "")
+                    console.print(f"{detail_prefix}  [yellow]➡[/yellow] {step_out_name}")
+
+                if not is_last:
+                    console.print("│")
 
             console.print()
 
         # Show final outputs (handle both dict and list formats)
```
aldbr: I guess this comment is not true anymore?
```suggestion
```

aldbr (changes requested):
Thanks for your contribution! 
I just have some minor comments 🙂

src/dirac_cwl/job/executor/__main__.py
amaltaro: Thank you for the review, @aldbr . 
Yes, directly accessing the object attributes should work as well. However, the `getattr` adds some protection in case the attribute changes for any reason (bug, non-backward compatible changes, etc).

What is the preferred behavior for this module:
a) have a clear exception being raised in case one of the expected attributes is not to be found?
b) or to gracefully deal with this with getattr and default values?

Given that this is already in a try/except block, I think option a) (which is your suggestion) is better. But better to get a confirmation please.

ryuwd: We discussed and we think the first option is the way to go

amaltaro:
@ryuwd @aldbr I applied the changes requested during review. 

However, one of the CI tests fails with an error unrelated to this PR:
https://github.com/DIRACGrid/dirac-cwl/actions/runs/30002435046/job/89190520820?pr=130

It looks like another PR that has recently been merged is not failing these unit tests (commit message length). Please let me know how to proceed here and/or how to retrigger CI tests.

aldbr (approved):

